PR #2, opened 2018-10-11
Added SPI mock with transaction based expectations

ryankurte:
Hey there,

Thanks for making a neat thing! I added an SPI mock with transaction based expectations so that it works with more complex APIs (see also: https://github.com/rust-embedded/embedded-hal/issues/94 and https://github.com/rust-embedded/embedded-hal/issues/64). It does use vectors heavily, so is never going to be `no_std`, but I think it's pretty useful.

Let me know if you need any changes / I would also like to propose moving the I2C mock to this model.

Thanks,

Ryan

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ to hardware.
 ## Status
 
 - [x] Simple I2C implementation
+- [x] Transactional SPI implementation
 - [x] No-op Delay implementation
 
 ## no\_std
@@ -40,6 +41,32 @@ let buf = [1, 2, 4];
 i2c.write(42, &buf).unwrap();
 assert_eq!(i2c.get_last_address(), Some(42));
 assert_eq!(i2c.get_write_data(), &[1, 2, 4]);
+```
+
+### SPI
+
+```rust
+use hal::blocking::spi::{Transfer, Write};
+use embedded_hal_mock::SPIMock;
+
+let mut spi = SPIMock::new();
+
+// Configure expectations
+spi.expect(vec![
+    Transaction::write(vec![1u8, 2u8]),
+    Transaction::transfer(vec![3u8, 4u8], vec![5u8, 6u8]),
+]);
+
+// Writing
+spi.write(&vec![1u8, 2u8]).unwrap();
+
+// Transferring
+let mut v = vec![3u8, 4u8];
+spi.transfer(&mut v).unwrap();
+assert_eq!(v, vec![5u8, 6u8]);
+
+// Finalise expectations
+spi.done();
 ```
 
 ### Delay
--- a/README.md
+++ b/README.md
@@ -47,14 +47,14 @@ assert_eq!(i2c.get_write_data(), &[1, 2, 4]);
 
 ```rust
 use hal::blocking::spi::{Transfer, Write};
-use embedded_hal_mock::{SPIMock, SPITransaction};
+use embedded_hal_mock::{SpiMock, SpiTransaction};
 
-let mut spi = SPIMock::new();
+let mut spi = SpiMock::new();
 
 // Configure expectations
 spi.expect(vec![
-    SPITransaction::write(vec![1u8, 2u8]),
-    SPITransaction::transfer(vec![3u8, 4u8], vec![5u8, 6u8]),
+    SpiTransaction::write(vec![1u8, 2u8]),
+    SpiTransaction::transfer(vec![3u8, 4u8], vec![5u8, 6u8]),
 ]);
 
 // Writing
--- a/README.md
+++ b/README.md
@@ -47,14 +47,14 @@ assert_eq!(i2c.get_write_data(), &[1, 2, 4]);
 
 ```rust
 use hal::blocking::spi::{Transfer, Write};
-use embedded_hal_mock::SPIMock;
+use embedded_hal_mock::{SPIMock, SPITransaction};
 
 let mut spi = SPIMock::new();
 
 // Configure expectations
 spi.expect(vec![
-    Transaction::write(vec![1u8, 2u8]),
-    Transaction::transfer(vec![3u8, 4u8], vec![5u8, 6u8]),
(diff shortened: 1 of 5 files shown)


src/spi.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Create a new mock SPI interface
+    pub fn new() -> Self {
+        SPIMock { expected: Arc::new(Mutex::new(Vec::new())) }
+    }
+
+    /// Set expectations on the SPI interface
+    /// This is a list of SPI transactions to be executed
+    pub fn expect(&mut self, expected: Vec<Transaction>) {
+        let mut e = self.expected.lock().unwrap();
+        assert_eq!(e.len(), 0);
```
dbrgn: I'm not sure if setting expectations should implicitly check previous state... I would expect the `expect` method to simply overwrite the previous expectations.

Maybe an `add_expect(&mut self, expected: Transaction)` might be interesting as an `append-this-to-current-expectations` kind of method?

ryankurte:
Thanks for all the feedback!

> If I understand your code correctly, the transactions are both used to specify data that will be returned, as well as asserting that commands are run in a certain order, right?

Pretty much, yeah. SPI is duplex, so you have both read and write at the same time. However, sometimes you don't care about the response, hence the delineation between `transfer` which writes and reads and `write` which only writes. For the write case, we can just check the data written, and in the transfer case, we also need to be able to fake the data read back so we can mock out drivers that care about the responses. 

So I modelled those as transactions, and combining them into a list lets a single driver method cause more than one SPI transaction, and yeah check the correctness and order of each, which is roughly in line with the most popular approach we've been discussing for the multiple SPI transactions in the hal https://github.com/rust-embedded/embedded-hal/issues/94#issuecomment-423787789.

> Another approach could be to not panic in the code itself, but instead to record the actual transactions. 

So, I actually wrote it that way initially, but on testing I decided it's more useful to cause a failure at the point the error exists. If you imagine a complex driver with a reasonable list of transactions, a diff of the transactions at the end is neat to look at but very hard to trace back to the call point that caused it. That's also why I went with assertions instead of result types, because the stacktrace thrown should immediately identify the particular transaction at fault.

If you feel strongly that you would prefer that approach, I can quickly re-implement it.

> The check method however could return a Result. The error case could also contain both the expected and the actual transaction.

The check method is only there to make sure all expectations have been met, there isn't really anything to compare at that point. I tried to implement it as a drop but that caused instruction faults and I didn't want to go further down that pathway.

> I'm not sure if setting expectations should implicitly check previous state... I would expect the expect method to simply overwrite the previous expectations.

Again, I started with that but concluded that overwriting a list of expectations would be a violation of those expectations. Happy to change that if you'd prefer, just needs to be documented either way.

> Could you add an example to README, and also update the "Status" section there?

Sure can.

dbrgn:
> but on testing I decided it's more useful to cause a failure at the point the error exists.

Yep, I think I agree :slightly_smiling_face: If we get a lot of feedback that the other approach is desired, this could still be changed (or be done as a second implementation).

> The check method is only there to make sure all expectations have been met, there isn't really anything to compare at that point.

Maybe add some function docs along the lines of "Assert that all expectations were met"? And maybe a name like `assertDone` or `ensureDone` might be better?

> Again, I started with that but concluded that overwriting a list of expectations would be a violation of those expectations.

I would not mix up setting expectations and checking expectations. While we could add a `strict` parameter to let the user decide, I think `spi.check(); spi.expect(new_expectations);` is much clearer.

In some cases you might want to re-use an SPI instance without handling old expectations. Right now that's not really possible.

Alternatively we could log a warning in `expect()` if some expectations were not met, and provide a `reset()` method to make resetting explicit.

ryankurte:
Fixed a bug with the expectation ordering (and added a test) and, I think, resolved all the suggestions ^_^

src/spi.rs
```diff
@@ -0,0 +1,222 @@
...
+        let w = self
+            .expected
+            .lock()
+            .unwrap()
+            .pop_front()
+            .expect("no expectation for spi::transfer call");
+        assert_eq!(w.expected_mode, Mode::Transfer);
+        assert_eq!(&w.expected_data, &buffer);
+        for i in 0..w.response.len() {
+            buffer[i] = w.response[i];
```
dbrgn: How do you know that the buffer has the same size as the response? This isn't guaranteed by the SPI trait, right?

Furthermore, is it expected that the response bytes are written to the buffer? It seems that in the [trait](https://docs.rs/embedded-hal/0.2.1/embedded_hal/blocking/spi/trait.Transfer.html#tymethod.transfer) the return reference lifetime is the same as the input buffer lifetime, so that's probably the only way for an SPI implementation to work? (Unfortunately the trait docs don't say anything about that...)

ryankurte: Fair point, it is inherent to the way SPI works, so I have added a length check to the transfer function.

It is expected / is the only way it works afaik.

README.md
```diff
@@ -42,6 +43,32 @@ assert_eq!(i2c.get_last_address(), Some(42));
...
+    Transaction::transfer(vec![3u8, 4u8], vec![5u8, 6u8]),
+]);
+
+// Writing
+spi.write(&vec![1u8, 2u8]).unwrap();
+
+// Transferring
+let mut v = vec![3u8, 4u8];
+spi.transfer(&mut v).unwrap();
+assert_eq!(v, vec![5u8, 6u8]);
```
dbrgn: Last change request (sorry): I'd call this variable `buf` to indicate that it's a buffer and not "just" a data vector :slightly_smiling_face:

After that I think it should be good to merge.

ryankurte: No worries, and done. Are you attached to exposing `SPIMock` at the top level? Because it might be nicer to have `spi::Mock` and `spi::Transaction` imo.

I could rename them then re-export with different names actually, maybe that's better.

dbrgn:
Thanks a lot, @ryankurte!

PR #4, opened 2018-10-13
Updated i2c::Mock to use transaction based expectations

ryankurte:
I'm not sure how you feel about breaking changes to the existing mocks, but I rebuild I2C to work along the same lines as the SPI mocks just merged.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -43,7 +43,7 @@ pub enum Mode {
 }
 
 /// I2C Transaction type
-/// 
+///
 /// Models an I2C read or write
 #[derive(Clone, Debug, PartialEq)]
 pub struct Transaction {
@@ -86,7 +86,7 @@ impl Transaction {
 }
 
 /// Mock I2C implementation
-/// 
+///
 /// This supports the specification and evaluation of expectations to allow automated testing of I2C based drivers.
 /// Mismatches between expectations will cause runtime assertions to assist in locating the source of the fault.
 pub type Mock<'a> = Generic<'a, Transaction>;
@@ -112,7 +112,7 @@ impl <'a>i2c::Write for Mock<'a> {
 
     fn write(&mut self, address: u8, bytes: &[u8]) -> Result<(), Self::Error> {
         let w = self.next().expect("no pending expectation for i2c::read call");
-            
+
         assert_eq!(w.expected_mode, Mode::Write, "i2c::write unexpected mode");
         assert_eq!(w.expected_addr, address, "i2c::write address mismatch");
         assert_eq!(w.expected_data, bytes, "i2c::write data does not match expectation");
@@ -131,7 +131,7 @@ impl <'a>i2c::WriteRead for Mock<'a> {
         buffer: &mut [u8],
     ) -> Result<(), Self::Error> {
         let w = self.next().expect("no pending expectation for i2c::read call");
-            
+
         assert_eq!(w.expected_mode, Mode::WriteRead, "i2c::write_read unexpected mode");
         assert_eq!(w.expected_addr, address, "i2c::write_read address mismatch");
         assert_eq!(w.expected_data, bytes, "i2c::write_read write data does not match expectation");
@@ -154,7 +154,6 @@ mod test {
         let expectations = [Transaction::write(0xaa, vec![10, 12])];
         let mut i2c = Mock::new(&expectations);
 
-
         i2c.write(0xaa, &vec![10, 12]).unwrap();
 
         i2c.done();
@@ -164,7 +163,6 @@ mod test {
     fn test_i2c_mock_read() {
         let expectations = [Transaction::read(0xaa, vec![1, 2])];
         let mut i2c = Mock::new(&expectations);
-
 
         let mut buff = vec![0u8; 2];
         i2c.read(0xaa, &mut buff).unwrap();
@@ -269,5 +267,28 @@ mod test {
         i2c.write(0xaa, &vec![10, 12]).unwrap();
 
         i2c.done();
+    }
+
+    #[test]
+    fn test_i2c_mock_cloned_done_ok() {
+        let expectations = [Transaction::read(0xaa, vec![1, 2])];
+        let mut i2c = Mock::new(&expectations);
+        let mut i2c_clone = i2c.clone();
+
+        let mut buff = vec![0u8; 2];
+        i2c.read(0xaa, &mut buff).unwrap();
+        assert_eq!(vec![1, 2], buff);
+
+        i2c.done();
+        i2c_clone.done();
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_i2c_mock_cloned_done_error() {
+        let expectations = [Transaction::read(0xaa, vec![1, 2])];
+        let i2c = Mock::new(&expectations);
+        let mut i2c_clone = i2c.clone();
+        i2c_clone.done();
     }
 }
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -10,13 +10,12 @@
 //! use embedded_hal::blocking::i2c::{Read, Write, WriteRead};
 //! use embedded_hal_mock::i2c::{Mock as I2cMock, Transaction as I2cTransaction};
 //!
-//! let mut i2c = I2cMock::new();
-//!
 //! // Configure expectations
-//! i2c.expect(vec![
+//! let expectations = [
 //!     I2cTransaction::write(0xaa, vec![1, 2]),
 //!     I2cTransaction::read(0xbb, vec![3, 4]),
-//! ]);
+//! ];
+//! let mut i2c = I2cMock::new(&expectations);
 //!
 //! // Writing
 //! i2c.write(0xaa, &vec![1, 2]).unwrap();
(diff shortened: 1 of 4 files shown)


ryankurte:
@eldruin neat! I hope to challenge you for that title :-P

My feeling is that for unit tests I am likely to be using a std environment anyway. That said, I originally used vectors to juggle expectations for async validation, now that all happens in line it'd be reasonable to replace them with arrays and indexes when someone has a demonstrated need for it.

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+/// I2C Transaction modes
+#[derive(Clone, Debug, PartialEq)]
+pub enum Mode {
+    Write,
+    Read,
+    WriteRead,
+}
+
+/// I2C Transaction type
+/// Models an I2C read or write
```
dbrgn: Please keep a space between the "headline" and the description text of docstrings. Otherwise rustdoc will render all that even in the overview :slightly_smiling_face:

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+            expected_addr: addr,
+            expected_data: expected,
+            response_data: response,
+        }
+    }
+}
+
+/// Mock I2C implementation
+/// This supports the specification and evaluation of expectations to allow automated testing of I2C based drivers.
+/// Mismatches between expectations will cause runtime assertions to assist in locating the source of the fault.
```
dbrgn: Same here (empty line missing).

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+}
+
+/// I2C Transaction type
+/// Models an I2C read or write
+#[derive(Clone, Debug, PartialEq)]
+pub struct Transaction {
+    expected_mode: Mode,
+    expected_addr: u8,
+    expected_data: Vec<u8>,
+    response_data: Vec<u8>,
```
dbrgn: I think using an `Option<Vec<u8>>` would be cleaner than creating an empty vector just to denote absence.

Or even this:

```rust
pub enum Mode {
    Write { expected: Vec<u8> },
    Read { response: Vec<u8> },
    WriteRead { expected: Vec<u8>, response: Vec<u8> },
}
```

...which would not allow invalid states to exist at all. (In this case it might be a good idea to rename `Mode` to `Expectation`.)

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+            expected: Arc::new(Mutex::new(VecDeque::new())),
         }
     }
 
-    /// Set data that will be read by `read()`.
-    pub fn set_read_data(&mut self, data: &'a [u8]) {
-        self.data = data;
+    /// Set expectations on the I2C interface
+    /// This is a list of I2C transactions to be executed in order
+    /// Note that setting this will overwrite any existing expectations
```
dbrgn: Same here (empty line missing).

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+        assert_eq!(expected.len(), 0);
     }
+}
 
-    /// Return the address that was used by the last read or write command.
-    pub fn get_last_address(&self) -> Option<u8> {
-        self.address
+impl Clone for Mock {
+    fn clone(&self) -> Mock {
+        Mock{expected: self.expected.clone()}
```
dbrgn: Personal preference, but I'd put spaces around the curly braces. I think the default rustfmt style agrees with that (although I don't generally apply rustfmt since it has some quirks when formatting more complex code).

Same for the SPI clone implementation.

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+            .expected
+            .lock()
+            .unwrap()
+            .pop_front()
+            .expect("no pending expectation for i2c::read call");
+
+        assert_eq!(w.expected_mode, Mode::Read);
+        assert_eq!(w.expected_addr, address);
+
+        assert_eq!(buffer.len(), w.response_data.len(), "mismatched response length for i2c:read");
```
dbrgn: I'm not sure if this should be enforced strictly, since a driver might want to read data partially. But it would make our transaction code more complicated. It's probably best to leave this line in, and to change it if someone shows a need for that.

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+        for i in 0..buffer.len() {
+            buffer[i] = w.response_data[i];
+        }
```
dbrgn: I think you can use `buffer.copy_from_slice(&w.response_data)` instead of the loop. Can probably be applied in the SPI code too. Also in the `WriteRead` implementation.

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
+    #[test]
+    fn test_i2c_mock_multiple() {
         let mut i2c = Mock::new();
-        let mut buf = [0; 3];
-        i2c.read(0, &mut buf).unwrap();
-        assert_eq!(buf, [0; 3]);
+
+        i2c.expect(vec![
+            Transaction::write(0xaa, vec![1u8, 2u8]),
+            Transaction::read(0xbb, vec![3u8, 4u8]),
```
dbrgn: Btw, I don't think you need to specify the `u8` suffix since Rust should be able to determine that from the `Transaction` type.

src/i2c.rs
```diff
@@ -7,186 +7,326 @@
...
         }
     }
 
-    /// Set data that will be read by `read()`.
-    pub fn set_read_data(&mut self, data: &'a [u8]) {
-        self.data = data;
+    /// Set expectations on the I2C interface
+    /// This is a list of I2C transactions to be executed in order
+    /// Note that setting this will overwrite any existing expectations
+    pub fn expect(&mut self, expected: Vec<Transaction>) {
```
dbrgn: At least here you could easily change the `expected` type to `&[Transaction]`, allowing the user to specify a static array.

dbrgn:
> My feeling is that for unit tests I am likely to be using a std environment anyway. That said, I originally used vectors to juggle expectations for async validation, now that all happens in line it'd be reasonable to replace them with arrays and indexes when someone has a demonstrated need for it.

In the `sgp30` driver, I think the `tests` module inherits the no_std property of the parent module: https://docs.rs/sgp30/0.2.0/src/sgp30/lib.rs.html#629 At least the `vec!` macro cannot be found in the tests.

Could you maybe attempt to at least take the vectors out of the public interface? I think that should be doable. Will require some lifetime annotations on the structs though (although we *could* make them all `&' static`, but I don't think that will work for all cases).

dbrgn:
Also, from testing a bit with my driver, I think it would be very helpful to add panic messages to the `assert_eq!` calls:

```
---- tests::get_feature_set stdout ----
thread 'tests::get_feature_set' panicked at 'assertion failed: `(left == right)`
  left: `[32, 47]`,
 right: `[32, 3]`', /home/danilo/.cargo/git/checkouts/embedded-hal-mock-a16b94889fef6415/c3562a3/src/i2c.rs:161:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Hard to say what the error above means, without generating a backtrace.

And another idea (could be implemented later): Sometimes I only want to test the processing of returned data, but don't care about the written data. Maybe we could add a placeholder type for "any data" that can be used in a write transaction.

src/i2c.rs
ryankurte: I wrote it that way originally, but, It seemed simpler to check `mode==mode` than to write a match cases to handle and unwrap the right enum combinations, and this way supported the lazy evaluation of expectations when I was trying that.

src/i2c.rs
ryankurte: That would not be valid with the current approach regardless, you'd have to set it as an expectation. Agreed it should be revisited if there is a need.

src/i2c.rs
ryankurte: Yep, can do

src/i2c.rs
dbrgn: I leave the decision to you :slightly_smiling_face:

ryankurte:
> Could you maybe attempt to at least take the vectors out of the public interface? I think that should be doable. Will require some lifetime annotations on the structs though (although we could make them all &' static, but I don't think that will work for all cases).

I generalised the common mock and made the interface take static arrays, probably needs some more doc updates but out of time for now.

dbrgn:
Squashed your commits. I also added two tests for the `clone` implementation. I didn't even realize at first what that enables, very nice idea :slightly_smiling_face:

dbrgn:
@ryankurte Regarding `Cow`, I played around for a while but could only come up with this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=46be8ea3a3694347d80c909c9f686f63

While this allows to pass owned values (e.g. a `Vec<T>`) directly into the mock by moving it, it makes ergonomics worse for references because the `[..]` syntax needs to be used...

Not sure if there's a better solution for this than the one we have right now.

ryankurte:
Yeah, I realised I couldn't implement anything without clone.
To pass slices and arrays (and anything else I guess) could the generic type instead require [Iterator<T>](https://doc.rust-lang.org/std/iter/trait.Iterator.html)?

dbrgn:
I wondered so too... Since you need to know the size of the iterable, you'd probably need `ExactSizeIterator<T>` though: https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

ryankurte:
Hmm, neither. I feel like there should be some neat approach with `where: ` but I can't see it yet, something for the future I guess.

Issue #1, opened 2018-09-30
Make this crate compatible with embedded-hal master


eldruin:
Hi,
I ran into a compilation problem when using this crate with the `embedded-hal` master version.
In my [pcf857x crate](https://github.com/eldruin/pcf857x-rs), if changing the `embedded-hal` dependency in the `Cargo.toml` file to:
```toml
embedded-hal = { git = "https://github.com/rust-embedded/embedded-hal" }
```
The tests cannot be run as the compiler complains that the `prelude::Read` and `prelude::Write` trait bounds were not satisfied.
Am I doing something wrong or is this crate incompatible with `embedded-hal` master?
There were some changes to [`embedded-hal`'s prelude](https://github.com/rust-embedded/embedded-hal/compare/v0.2.1...HEAD#diff-a7dd294c4d03f040db2403cd577865d7). Could this be the cause?

PS: Thanks for this crate! I use it all the time :)

dbrgn:
Hi. I'm just back from vacations. I need a few days to clear my backlog and will revisit this issue later on.

At first glance, the problem is that your project depends on embedded-hal@master while the mock crate depends on embedded-hal@0.2. Even if the traits are compatible feature-wise, different versions are incompatible by design. I think a feature flag that uses master instead of a released version would help, although I'm not sure if that can be released to crates.io.

dbrgn:
I don't think crates with dependency to non-published crates can be published.

I'm not sure what the best way to implement this would be. Maybe I'll maintain a `next` branch that points to master and that can be used directly from `Cargo.toml` using a git dependency.

ryankurte:
Will this still be an issue when we've released embedded-hal v0.3.0 and updated this to point at it?

dbrgn:
As long as people use the `master` version of `embedded-hal`, I think yes.

The dependency tree looks like this:

```
a
|- embedded-hal@master
|- embedded-hal-mock@0.4.0
   |- embedded-hal@0.3.0
```

This means that there are two versions of embedded-hal. The traits implemented by embedded-hal-mock for embedded-hal@0.3.0 are not compatible with the traits provided by embedded-hal@master, even if their source code is identical.

Of course if both @eldruin and we use the released embedded-hal 0.3.0, then it's going to work.

eldruin:
> The traits implemented by embedded-hal-mock for embedded-hal@0.3.0 are not compatible with the traits provided by embedded-hal@master, even if their source code is identical.

[...]

PR #10, opened 2018-11-28
Updated using latest instructions with cargo patch

ryankurte:
Per #1 swaps the instructions to use cargo patch functionality


dbrgn:
Thanks!

PR #12, opened 2018-12-26
Fix error messages for unfulfilled expectations

### LINKED ISSUE #11: The error messages for un-fulfilled `write` and `write_read` calls is incorrect
They say `read call` instead of their respective actual calls:

https://github.com/dbrgn/embedded-hal-mock/blob/master/src/i2c.rs#L117

https://github.com/dbrgn/embedded-hal-mock/blob/master/src/i2c.rs#L136

Looks like a copy-pasta error :)

jasonpeacock:
Closes #11


dbrgn:
0.4.1 is released with this fix.

PR #14, opened 2018-12-31
Spidev added spi FullDuplex

simonsso:
Happy New Year!

Added implementation for single byte send/read via the non blocking embedded_hal::spi::FullDuplex trait.

Limitations: Unit test must be updated when changing from a implementation based on read() and write() calls to one with transfer calls as they do not share data.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -33,8 +33,9 @@
 //! // Finalise expectations
 //! spi.done();
 //! ```
-
+extern crate nb;
 use hal::blocking::spi;
+use hal::spi::FullDuplex;
 
 use common::Generic;
 use error::MockError;
@@ -46,6 +47,10 @@ pub enum Mode {
     Write,
     /// Write and read transaction
     Transfer,
+    /// Send transaction
+    Send,
+    /// After a send transaction in real HW a Read is available
+    Read,
 }
 
 /// SPI transaction type
@@ -73,7 +78,23 @@ impl Transaction {
         Transaction {
             expected_mode: Mode::Transfer,
             expected_data: expected,
-            response,
+            response: response,
+        }
+    }
+    /// Create a transfer transaction
+    pub fn send(expected: u8 ) -> Transaction {
+        Transaction {
+            expected_mode: Mode::Send,
+            expected_data: [expected].to_vec(),
+            response: Vec::new(),
+        }
+    }
+    /// Create a transfer transaction
+    pub fn read(response: u8) -> Transaction {
+        Transaction {
+            expected_mode: Mode::Read,
+            expected_data: Vec::new(),
+            response: [response].to_vec(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


dbrgn:
Thanks for the contribution!

Unfortunately I'm not very familiar with the `spi::FullDuplex` trait. @ryankurte have you used it in practice? Maybe you could do a quick review?

Regarding this:

> Limitations: Unit test must be updated when changing from a implementation based on read() and write() calls to one with transfer calls as they do not share data.

I think we should try to encode this fact into the type system. I wonder whether we should have separate mock types for blocking and non-blocking modes?

simonsso:
SPI reads and send data on the same clock generated from master. In a real device you could run half a transaction chunk with send and read (as in FullDuplex trait) and then for the other half,  use transfer, SPI does allow for breaks between words. Already with the existing code, if implementation split one data transfer into two separate function calls, unit test must be rewritten too. This patch is compatible with this and have added transaction types for the single byte transfers in the enum.

dbrgn:
Thanks for the review, @eldruin!

Regarding the formatting, I think I'll add a rustfmt setup once this is merged, and will then reformat all existing code so it's consistent.

src/spi.rs
```diff
@@ -102,6 +123,31 @@ impl <'a>spi::Write<u8> for Mock<'a> {
...
+        let data = self.next().expect("no expectation for spi::send call");
+        assert_eq!(data.expected_mode, Mode::Send, "spi::send unexpected mode");
+        assert_eq!(data.expected_data[0], buffer, "spi::send data does not match expectation");
+        Ok(())
+    }
+    /// spi::FullDuplex implementeation for Mock
+    ///
+    /// This will call the nonblocking read/write primitives.
+
+    fn read(&mut self) -> nb::Result< u8, Self::Error> {
```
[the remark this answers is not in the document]
simonsso: I added a few test cases below, do is there any other place you want it at?

src/spi.rs
```diff
@@ -13,15 +13,25 @@
 //!
 //! use embedded_hal::blocking::spi::{Transfer, Write};
 //! use embedded_hal_mock::spi::{Mock as SpiMock, Transaction as SpiTransaction};
-//!
+//! use embedded_hal::spi::FullDuplex;
+//! 
 //! // Configure expectations
 //! let expectations = [
+//!     SpiTransaction::send(0x09),
```
eldruin: Thanks for adding the example, however I would add this `FullDuplex`example in addition (separately) to the original example as I would say that they are separate concepts and they would be easier to understand separately.

simonsso: They are not separate concepts, it it possible to interchange freely between using Fullduplex send/read-pairs and transfer. I wanted to highlight this from the example, feel free to edit the example to your liking there is code below in the tests section you could cut and paste from.

ryankurte:
Ooh, neat!

> Limitations: Unit test must be updated when changing from a implementation based on read() and write() calls to one with transfer calls as they do not share data.

I'm not sure I understand this? `read` and `write` are implemented on the SpiTransaction enum so they're interchangeable as far as I can see?

> I think we should try to encode this fact into the type system. I wonder whether we should have separate mock types for blocking and non-blocking modes?

I agree that they're not necessarily separate / can be used interchangeably, so I think this is the way to go (also the implementations are effectively namespaced by the traits the implement).

dbrgn:
> I agree that they're not necessarily separate / can be used interchangeably, so I think this is the way to go (also the implementations are effectively namespaced by the traits the implement).

I'm not sure I understand you correctly, are you suggesting separate mock types, or a single type that implements both traits?

simonsso:
Sorry if I have confused you. I try to clarify my reasoning: SPI transaction can be implemented with either FullDuplex send/read or with Transfer transfer. But on the physical layer the interface is just shifting bits. On 8 clocks one byte is written and one bytes is read. 

So the software could split a 1000 byte transaction into two 500 byte transactions and have the same physical behavior. (Or even1000 one-byte-tranasctions) With current implementation if a user refactors from one 1000 byte transaction to two 500 byte transactions, the mock  must also be changed. This is true for the FullDuplex trait too.

No I do not suggest to change this structure, that is why I have added items in the same enum and kept the same structure. 

Even as a send/read implementation will be very verbose, it would probably be more confusing to have one huge stream of data when a developer creating a unit test thinks in transactions.

dbrgn:
Ok, as written before I don't have practical experience with the `HalfDuplex` mode, but I think the PR looks good now :slightly_smiling_face: If someone has improvements, further PRs are welcome.

Thanks @simonsso!

PR #18, opened 2019-02-05
Mock Pin impl

ryankurte:
Mock pin implementation, this implements mocks for the digital::v2 traits and is thus blocked on an embedded-hal release including these.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,4 +22,3 @@ edition = "2018"
 [dependencies]
 embedded-hal = { version = "0.2.3", features = ["unproven"] }
 nb = "0.1.1"
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,6 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = { version = "0.2.2", features = ["unproven"] }
+embedded-hal = { version = "0.2.3-rc.1", features = ["unproven"] }
 nb = "0.1.1"
-
-[patch.crates-io]
-embedded-hal = { git = "https://github.com/rust-embedded/embedded-hal" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,9 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = "0.2"
+embedded-hal = { version = "0.2.2", features = ["unproven"] }
 nb = "0.1.1"
+
+[patch.crates-io]
+embedded-hal = { git = "https://github.com/rust-embedded/embedded-hal" }
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,6 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = { version = "0.2.3-rc.1", features = ["unproven"] }
+embedded-hal = { version = "0.2.3", features = ["unproven"] }
 nb = "0.1.1"
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,14 +1,14 @@
 use std::io;
 
 /// Errors that may occur during mocking.
-#[derive(Debug)]
+#[derive(PartialEq, Clone, Debug)]
 pub enum MockError {
     /// An I/O-Error occurred
-    Io(io::Error),
+    Io(io::ErrorKind),
 }
 
 impl From<io::Error> for MockError {
     fn from(e: io::Error) -> Self {
-        MockError::Io(e)
+        MockError::Io(e.kind())
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,3 +25,4 @@ pub mod delay;
 pub mod i2c;
 pub mod serial;
 pub mod spi;
+pub mod pin;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,6 @@ pub use crate::error::MockError;
 pub mod common;
 pub mod delay;
 pub mod i2c;
+pub mod pin;
 pub mod serial;
 pub mod spi;
-pub mod pin;
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -97,10 +97,10 @@ impl TransactionKind {
 }
 
 /// Mock Pin implementation
-pub type Mock<'a> = Generic<'a, Transaction>;
+pub type Mock = Generic<Transaction>;
 
 /// Single digital push-pull output pin
-impl <'a>OutputPin for Mock<'a> {
+impl OutputPin for Mock {
     /// Error type
     type Error = MockError;
 
@@ -108,7 +108,7 @@ impl <'a>OutputPin for Mock<'a> {
     fn set_low(&mut self) -> Result<(), Self::Error> {
         let Transaction{kind, err} = self.next().expect("no expectation for pin::set_low call");
 
-        assert_eq!(*kind, TransactionKind::Set(false), "expected pin::set_low");
+        assert_eq!(kind, TransactionKind::Set(false), "expected pin::set_low");
         
         match err {
             Some(e) => Err(e.clone()),
@@ -120,7 +120,7 @@ impl <'a>OutputPin for Mock<'a> {
     fn set_high(&mut self) -> Result<(), Self::Error> {
         let Transaction{kind, err} = self.next().expect("no expectation for pin::set_high call");
 
-        assert_eq!(*kind, TransactionKind::Set(true), "expected pin::set_high");
+        assert_eq!(kind, TransactionKind::Set(true), "expected pin::set_high");
         
         match err {
             Some(e) => Err(e.clone()),
@@ -129,7 +129,7 @@ impl <'a>OutputPin for Mock<'a> {
     }
 }
 
-impl <'a>InputPin for Mock<'a> {
+impl InputPin for Mock {
     /// Error type
     type Error = MockError;
 
@@ -144,7 +144,7 @@ impl <'a>InputPin for Mock<'a> {
         if let Some(e) = err { 
             Err(e.clone())
         } else if let TransactionKind::Get(v) = kind {
-            Ok(*v == true)
+            Ok(v == true)
         } else {
             unreachable!();
         }
@@ -161,7 +161,7 @@ impl <'a>InputPin for Mock<'a> {
         if let Some(e) = err { 
             Err(e.clone())
         } else if let TransactionKind::Get(v) = kind {
-            Ok(*v == false)
+            Ok(v == false)
         } else {
             unreachable!();
         }
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -1,0 +1,219 @@
+//! Mock digital InputPin and OutputPin implementations
+//! 
+//! ```
+//! use std::io::ErrorKind;
+//! 
+//! use embedded_hal_mock::MockError;
+//! use embedded_hal_mock::pin::{Transaction as PinTransaction, Mock as MockPin};
+//! use embedded_hal::digital::v2::{InputPin, OutputPin};
+//! 
+//! let err = MockError::Io(ErrorKind::NotConnected);
+//! 
+//! // Configure expectations
+//! let expectations = [
+//!     PinTransaction::get(true),
+//!     PinTransaction::get(true),
+//!     PinTransaction::set(false),
+//!     PinTransaction::set(true).with_error(err.clone()),
+//! ];
+//! 
+//! // Create pin
+//! let mut pin = MockPin::new(&expectations);
+//! 
+//! // Run and test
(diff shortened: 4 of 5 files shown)


Cargo.toml
```diff
@@ -20,5 +20,8 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = "0.2"
+embedded-hal = { version = "0.2.2", features = ["unproven"] }
 nb = "0.1.1"
+
+[patch.crates-io]
+embedded-hal = { path = "../embedded-hal" }
```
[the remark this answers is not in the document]
ryankurte: yeah it's impossible to compile until there's a new embedded-hal release, i can swap to git but the override is stuck til then.

src/pin.rs
```diff
@@ -0,0 +1,138 @@
...
+mod test {
+    use super::*;
+
+    use embedded_hal::digital::v2::{OutputPin, InputPin};
+    use TransactionKind::{Get, Set};
+
+    #[test]
+    fn test_input_pin() {
+        let expectations = [
+            Transaction::new(Get(true), None),
```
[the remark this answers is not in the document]
ryankurte: hmm, i could use a builder on the basis that the error case should be the least regularly used?

Cargo.toml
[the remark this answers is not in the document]
eldruin: Yes but git would make it possible to compile.

src/pin.rs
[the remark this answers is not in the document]
eldruin: `.with_error` sounds good.

src/error.rs
```diff
@@ -1,14 +1,14 @@
 use std::io;
 
 /// Errors that may occur during mocking.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum MockError {
     /// An I/O-Error occurred
-    Io(io::Error),
+    Io(io::ErrorKind),
```
dbrgn: :+1:

src/lib.rs
```diff
@@ -24,3 +24,6 @@ pub mod common;
 pub mod delay;
 pub mod i2c;
 pub mod spi;
+
+/// Mock InputPin and OutputPin implementations
```
dbrgn: I'd put mod documentation inside the module (with `//! comments`). Then you have a description of what the file does when you open it :slightly_smiling_face:

Also, another paragraph with a quick explanation on how to use this mock and a small example would be nice!

src/pin.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::common::Generic;
+use crate::error::MockError;
+
+use embedded_hal::digital::v2::{OutputPin, InputPin};
+
+/// MockPin transaction
+#[derive(Debug)]
+pub struct Transaction {
+    kind: TransactionKind,
+    err: Option<MockError>,
```
dbrgn: This is for "faking" an error during a transaction, right?

src/pin.rs
```diff
@@ -0,0 +1,138 @@
...
+        let mut s = self.clone();
+
+        let Transaction{kind, err} = s.next().expect("no expectation for pin::is_high call");
+
+        if let Some(e) = err { 
+            Err(e.clone())
+        } else if let TransactionKind::Get(v) = kind {
+            Ok(*v == true)
+        } else {
+            panic!("expected pin::is_high (actual: {:?}", kind);
```
dbrgn: A closing `)` is missing here.

src/pin.rs
```diff
@@ -0,0 +1,138 @@
...
+        let mut s = self.clone();
+
+        let Transaction{kind, err} = s.next().expect("no expectation for pin::is_low call");
+
+        if let Some(e) = err { 
+            Err(e.clone())
+        } else if let TransactionKind::Get(v) = kind {
+            Ok(*v == false)
+        } else {
+            panic!("expected pin::is_low (actual: {:?}", kind);
```
dbrgn: A closing `)` is missing here.

src/pin.rs
dbrgn: Why the check-and-panic here and an assertion for the `set_` methods? Maybe the same approach could be used for both? I think the error messages differ otherwise, which might confuse users.

src/pin.rs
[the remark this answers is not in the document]
dbrgn: I agree, `.with_error` sounds good.

src/pin.rs
ryankurte: yep, added some documentation and a test. i guess rustdocs would be a good addition too.

src/pin.rs
ryankurte: For the set case we can check for `Set(true)` or `Set(false)` for `set_high` and `set_low`, which makes the assert easy, for the get case `is_high` and `is_low` are both valid so I ended up with that. Open to suggestions if you have a better idea?

ryankurte:
hmm i'm not sure i understand why the build is failing, same idiom is fine in other modules and on my stable/nightly.

src/pin.rs
```diff
@@ -0,0 +1,203 @@
...
+//! ```
+
+
+use crate::common::Generic;
+use crate::error::MockError;
+
+use embedded_hal::digital::v2::{OutputPin, InputPin};
+
+/// MockPin transaction
+#[derive(Debug)]
```
eldruin: This should probably also implement `Clone` and `PartialEq`

src/pin.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// Create a new get transaction
+    pub fn set(value: bool) -> Transaction {
+        Transaction::new(TransactionKind::Set(value))
+    }
+
+    /// Add an error return to a transaction
+    /// This is used to mock failure behaviours
+    pub fn with_error(mut self, error: MockError) -> Self {
```
eldruin: I did not expect `mut` here but I guess it is fine.

src/pin.rs
```diff
@@ -0,0 +1,203 @@
...
+//! use embedded_hal_mock::MockError;
+//! use embedded_hal_mock::pin::{Transaction as PinTransaction, Mock as MockPin};
+//! use embedded_hal::digital::v2::{InputPin, OutputPin};
+//! 
+//! // Configure expectations
+//! let expectations = [
+//!     PinTransaction::get(true),
+//!     PinTransaction::get(true),
+//!     PinTransaction::set(false),
+//!     PinTransaction::set(true).with_error(MockError::Io(ErrorKind::NotConnected)),
```
eldruin: Do you think it would be possible to make this line somehow shorter?

dbrgn: We could impl `From<io::ErrorKind> for MockError` and use `into()`, but I think I'd actually prefer the explicit version.

ryankurte: Hmm, I just need an `Error` instance, it doesn't have to me a `MockError` it was just conveniently there.

I could swap to `io::Error::new(ErrorKind::Other, "whoops")` or something, but it's not measurably shorter.

ryankurte: Oh wait yes I do. Could make a `MockError::not_connected()` helper or something?

src/pin.rs
ryankurte: Added an `is_get` method and swapped to `assert` with `unreachable`... does that seem reasonable?

src/pin.rs
dbrgn: But then we'd have to add a helper for every possible `ErrorKind`, right? I doubt that this is a good general solution.

In real code, if you re-use the same error a lot, you can always create a `MockError` instance and simply clone it every time you need it:

```rust
let not_connected = MockError::Io(ErrorKind::NotConnected);
...
PinTransaction::set(true).with_error(not_connected.clone())
```

src/error.rs
```diff
@@ -1,14 +1,14 @@
 use std::io;
 
 /// Errors that may occur during mocking.
-#[derive(Debug)]
+#[derive(PartialEq, Clone, Debug)]
```
eldruin:
```suggestion
#[derive(PartialEq, Clone, Copy, Debug)]
```
Could we make the error types `Copy`? The types are trivial and furthermore [`std::io::ErrorKind`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#impl-Copy) also is.

src/pin.rs
eldruin: @dbrgn that solution is probably enough. I also added a suggestion of making the error types copyable (see below) which would further simplify this.

src/error.rs
dbrgn: Right now it is, but if we add more error variants in the future it might become non-copy.

(On the other hand it's just a testing framework, so that might be fine. I'm not sure what would be best.)

eldruin: I see. I can also live with the clone at least for now.

src/pin.rs
dbrgn: Thanks! Shouldn't docs be sufficiently covered by https://docs.rs/?

Cargo.toml
```diff
@@ -20,5 +20,9 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = "0.2"
+embedded-hal = { version = "0.2.2", features = ["unproven"] }
 nb = "0.1.1"
+
+[patch.crates-io]
+embedded-hal = { git = "https://github.com/rust-embedded/embedded-hal" }
```
dbrgn: Do you know when this will be released? I think the merge should probably wait until we don't have to patch dependencies with git master references.

src/pin.rs
```diff
@@ -0,0 +1,219 @@
+//! Mock digital InputPin and OutputPin implementations
```
dbrgn: Maybe link to embedded-hal docs?

```diff
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -1,4 +1,7 @@
-//! Mock digital InputPin and OutputPin implementations
+//! Mock digital [`InputPin`] and [`OutputPin`] implementations
+//!
+//! [`InputPin`]: https://docs.rs/embedded-hal/latest/embedded_hal/digital/trait.InputPin.html
+//! [`OutputPin`]: https://docs.rs/embedded-hal/latest/embedded_hal/digital/trait.OutputPin.html
```

(Unfortunately [RFC1946](https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html) is not yet stabilized...)

src/pin.rs
```diff
@@ -0,0 +1,219 @@
...
+//! 
+//! ```
+//! use std::io::ErrorKind;
+//! 
+//! use embedded_hal_mock::MockError;
+//! use embedded_hal_mock::pin::{Transaction as PinTransaction, Mock as MockPin};
```
dbrgn: Nitpicking: In other docs we always used the target type as prefix: `I2cMock` and `I2cTransaction`, `SpiMock` and `SpiTransaction`. Here we use `PinTransaction` but `MockPin`.

src/pin.rs
```diff
@@ -0,0 +1,219 @@
...
+//! use embedded_hal::digital::v2::{InputPin, OutputPin};
+//! 
+//! let err = MockError::Io(ErrorKind::NotConnected);
+//! 
+//! // Configure expectations
+//! let expectations = [
+//!     PinTransaction::get(true),
+//!     PinTransaction::get(true),
```
dbrgn: I wonder if it would be better to have explicit pin states: `enum PinState { High, Low }`.

It would look a bit less confusing when creating the list of expectations:

```rust
 let expectations = [
    PinTransaction::get(PinState::High),
    PinTransaction::get(PinState::High),
];

assert_eq!(pin.is_high().unwrap(), true);
assert_eq!(pin.is_low().unwrap(), false);
```

Cargo.toml
ryankurte: I think we're waiting on more thoughts and reviews on https://github.com/rust-embedded/embedded-hal/pull/127, at which point we should be able to push a minor release.
I bumped it for more reviews today, will follow up in a couple of days once anyone else has had a chance to comment.

src/pin.rs
ryankurte: sorry i meant doctests, which are now there too.

Cargo.toml
dbrgn: Sounds good!

ryankurte: we've pushed 0.2.3-rc.1 which contains the new digital traits, and i've updated this PR to point to it.
how do y'all feel about depending on an -rc crate?

eldruin: Would this play well with drivers depending on `embedded-hal = "0.2"` ? Would cargo pick the rc version for everybody or would we then need to do the semver trick thing?

ryankurte: it turns out rc versions are also cursed with incompatibility...

we've shipped 0.2.3 now and i've updated this to match so i think this is all resolved.

dbrgn:
Version 0.6.0 released including this PR.

PR #21, opened 2019-02-23
Add serial device mock

mciantyre:
I took a stab at implementing a mock for the HAL serial traits. The goal was to keep an interface that was consistent with the SPI and I2C APIs. We have serial transactions (such as read, write, and flush) that are passed into a `serial::Mock`. Like the SPI and I2C mocks, the serial mock panics if the operations on it do not match the listed transactions.

Limitations and workarounds:

- The `embedded_hal` crate has a blanket, default that implements the blocking serial `Write` trait in terms of the non-blocking variant. I could not find a way to override that blanket impl and manually implement `blocking::serial::Write`, so I simply adopted the default implementation. This means that the mock's `blocking::serial::Write` implementation is written in terms of the non-blocking variant. Put another way: we cannot distinguish between blocking and non-blocking writes with the mock. If anyone knows how to work around this, please let me know! This doesn't stop users from expecting blocking writes; they just need to do it in terms of a write transaction with multiple words.
- Given the limitation above, I had trouble adopting the `Generic` mock that's used in the I2C and SPI mocks. Since we cannot distinguish non-blocking writes from blocking writes, we needed a way to modify how transactions are placed into the queue. Specifically, we need a single "write" transaction with multiple bytes to convert into multiple "write" transactions that each expect a single byte. In order to keep a consistent API with the I2C and SPI mocks, something needed to give. I found it easier to implement a different transaction queue than modify the one shared across I2C and SPI mocks.

Quick demo:

```rust
use embedded_hal::serial::{Read, Write};
use embedded_hal_mock::serial::{
    Mock as SerialMock,
    Transaction as SerialTransaction,
};

// Configure expectations
let expectations = [
    SerialTransaction::read(0x0A),
    SerialTransaction::read(0xFF),
    SerialTransaction::write([1, 2]),
    SerialTransaction::flush(),
];

let mut serial = SerialMock::new(&expectations);

let word = serial.read().unwrap();
assert_eq!(word, 0x0A);
let word = serial.read().unwrap();
assert_eq!(word, 0xFF);

serial.write(1).unwrap();
serial.write(2).unwrap();

serial.flush().unwrap();
serial.done();
```

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -407,7 +407,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "called serial::write with no expectation")]
     fn test_serial_mock_blocking_write_more_than_expected() {
         let ts = [Transaction::write([0xAB, 0xCD])];
         let mut ser = Mock::new(&ts);
@@ -416,7 +416,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "unsatisfied expectations")]
     fn test_serial_mock_blocking_write_not_enough() {
         let ts = [Transaction::write([0xAB, 0xCD, 0xEF, 0x00])];
         let mut ser = Mock::new(&ts);
@@ -425,7 +425,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "serial::write expected to write")]
     fn test_serial_mock_wrong_write() {
         let ts = [Transaction::write([0x12])];
         let mut ser = Mock::new(&ts);
@@ -449,7 +449,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "unsatisfied expectations")]
     fn test_serial_mock_pending_transactions() {
         let ts = [Transaction::read(0x54)];
         let ser = Mock::new(&ts);
@@ -457,7 +457,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "expected to perform a serial transaction 'Read(")]
     fn test_serial_mock_expected_read() {
         let ts = [Transaction::read(0x54)];
         let mut ser = Mock::new(&ts);
@@ -465,7 +465,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "expected to perform a serial transaction 'Write(")]
     fn test_serial_mock_expected_write() {
         let ts = [Transaction::write([0x54])];
         let mut ser = Mock::new(&ts);
@@ -473,7 +473,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
+    #[should_panic(expected = "expected to perform a serial transaction 'Flush'")]
     fn test_serial_mock_expected_flush() {
         let ts = [Transaction::flush()];
         let mut ser: Mock<u128> = Mock::new(&ts);
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -3,8 +3,7 @@
 //! We may expect serial read and write transactions on a mock
 //! Serial device. Note that, in the `embedded_hal` crate, there
 //! are the non-blocking serial traits, and there is a blocking
-//! `serial::Write` trait that is defined in terms of the non-blocking
-//! variant. We may use the same mock to mock both interfaces.
+//! variant. We may use the same mock for both interfaces.
 //!
 //! ## Usage: non-blocking serial traits
 //!
@@ -239,22 +238,37 @@ pub struct Mock<'a, Word> {
 impl<'a, Word: Clone> Mock<'a, Word> {
     /// Create a serial mock that will expect the provided transactions
     pub fn new(transactions: &'a [Transaction<Word>]) -> Self {
-        let expected_modes = transactions
-            .iter()
-            .fold(VecDeque::new(), |mut modes, transaction| {
-                modes.extend(transaction.mode.clone());
-                modes
-            });
-        Mock {
-            expected_modes: Arc::new(Mutex::new(Some(expected_modes))),
+        let mut ser = Mock {
+            expected_modes: Arc::new(Mutex::new(None)),
             marker: PhantomData,
-        }
+        };
+        ser.expect(transactions);
+        ser
+    }
+
+    /// Set expectations on the interface
+    ///
+    /// This is a list of transactions to be executed in order.
+    /// Note that setting this will overwrite any existing expectations
+    pub fn expect(&mut self, transactions: &'a [Transaction<Word>]) {
+        let mut lock = self
+            .expected_modes
+            .lock()
+            .expect("unable to lock serial mock in call to expect");
+        *lock = Some(
+            transactions
+                .iter()
+                .fold(VecDeque::new(), |mut modes, transaction| {
+                    modes.extend(transaction.mode.clone());
+                    modes
+                }),
+        );
     }
 
     /// Consumes the mock, asserting that all exepctations were met
     pub fn done(self) {
         // Note that we take self by value, unlike the generic versions that
-        // take by reference. It wouldn't make sense to call done() twice anyway :P
+        // take by mut reference. It wouldn't make sense to call done() twice anyway :P
         let mut lock = self
             .expected_modes
             .lock()
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -112,7 +112,6 @@ use embedded_hal::serial;
 use crate::error::MockError;
 
 use std::collections::VecDeque;
-use std::marker::PhantomData;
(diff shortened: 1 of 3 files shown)


src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+//! let mut serial = SerialMock::new(&expectations);
+//!
+//! // Expect two reads
+//! let word = serial.read().unwrap();
+//! assert_eq!(word, 0x0A);
+//! let word = serial.read().unwrap();
+//! assert_eq!(word, 0xFF);
+//!
+//! // When designing against the non-blocking serial
+//! // trait, we expect two separate writes. These could be
```
dbrgn: could or should?

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+//! // expressed as two separate transactions, too. See (1) above.
+//! serial.write(1).unwrap();
+//! serial.write(2).unwrap();
+//!
+//! // Finally, we expect a flush
+//! serial.flush().unwrap();
+//!
+//! // When you believe there are no more calls on the mock,
+//! // you may call done(), or let the mock drop out without
+//! // any other checks
```
dbrgn: I think it's bad practice if you skip the `done()` call, as you won't notice if some expectations were not fulfilled.

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+//! // two words are written. See (2) above.
+//! serial.bwrite_all(&[1, 2]).unwrap();
+//!
+//! // Finally, we expect a flush. Note that this is
+//! // a *blocking* flush from the blocking serial trait.
+//! serial.bflush().unwrap();
+//!
+//! // When you believe there are no more calls on the mock,
+//! // you may call done(), or let the mock drop out without
+//! // any other checks
```
dbrgn: Same here, I think dropping the mock without calling `done()` is bad practice.

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+
+// This module is implemented a little differently than
+// the spi and i2c modules. We'll note that, unlike the
+// spi and i2c modules which share the foundational Generic
+// transaction queue, we provide our own implementation.
+// We found that, in keeping with the established API design
+// and the unique features of the embedded_hal serial traits
+// (described in the note below), this was a necessary trade-
+// off. We welcome any other ideas that allow us to take
+// advantage of the common components.
```
dbrgn: :+1: for describing your implementation/design choices :slightly_smiling_face:

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// so that we can take it in the call to `done()`.
+    expected_modes: Arc<Mutex<Option<VecDeque<Mode<Word>>>>>,
+    /// We'll note that this phantom type and lifetime is actually
+    /// unnecessary. As noted above, we try to keep a consistent
+    /// public API with the existing spi and i2c modules. Given that
+    /// this lifetime is necessary for them, it is also necessary here.
+    ///
+    /// Of course, this could quickly be removed if the overall public
+    /// API changes.
+    marker: PhantomData<&'a Word>,
```
dbrgn: Hm, what do you mean with "Given that this lifetime is necessary for them, it is also necessary here"?

If the marker / lifetime is not required, then you can remove it. However the lifetime is used below, right?

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+                    modes.extend(transaction.mode.clone());
+                    modes
+                }),
+        );
+    }
+
+    /// Consumes the mock, asserting that all exepctations were met
+    pub fn done(self) {
+        // Note that we take self by value, unlike the generic versions that
+        // take by mut reference. It wouldn't make sense to call done() twice anyway :P
```
dbrgn: Wouldn't re-using a mock instance be a valid use case?

For example:

```rust
mock.expect(...);
// do something with the mock
mock.done();
mock.expect(...);
// do some other things with the mock
mock.done();
```

For that to work, done would need to take a non-consuming reference to self.

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+        }
+    }
+}
+
+// Note: we attempted to provide our own implementation of
+// embedded_hal::blocking::serial::Write. However, we're unable
+// to override it due to the blanket default implementation provided by
+// the embedded_hal crate. It comes down to the fact that, if we were
+// to provide an embedded_hal::blocking::serial::Write implementation
+// here, any user of embedded_hal would be free to implement the *defuault*
```
dbrgn:
```suggestion
// here, any user of embedded_hal would be free to implement the *default*
```

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+            Mode::Flush => Ok(()),
+            mode => panic!(
+                "expected to perform a serial transaction '{:?}' but instead did a flush",
+                mode
+            ),
+        }
+    }
+}
+
+// Note: we attempted to provide our own implementation of
```
dbrgn:
```suggestion
// Note: We attempted to provide our own implementation of
```

src/serial.rs
```diff
@@ -0,0 +1,484 @@
...
+    #[test]
+    fn test_serial_mock_blocking_write() {
+        let ts = [Transaction::write([0xAB, 0xCD, 0xEF])];
+        let mut ser = Mock::new(&ts);
+        ser.bwrite_all(&[0xAB, 0xCD, 0xEF]).unwrap();
+        ser.done();
+    }
+
+    #[test]
+    #[should_panic]
```
dbrgn: You should be able to assert a certain panic message. Example:

> #[should_panic(expected = "Divide result is zero")]

See https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#testing-panics

src/serial.rs
mciantyre: Good call; updating the docs to suggest calling `done()` in order to assert no pending transactions.

src/serial.rs
mciantyre: Could is correct. Given the nuances of the implementation, either of the expectations

```
let batch = [
    SerialTransaction::write([1, 2]),
];

let serial = [
    SerialTransaction::write([1]),
    SerialTransaction::write([2]),
];
```
could satisfy the following serial transactions:

```
// Mathes 'batch' or 'serial' expectations
serial.bwrite_all(&[1, 2]).unwrap();

// Also matches 'batch' or 'serial' expectations
serial.write(1).unwrap();
serial.write(2).unwrap();
```

The comment tries to indicate that the user could write their expectations resembling

```
let expectations = [
    SerialTransaction::read(0x0A),
    SerialTransaction::read(0xFF),
    SerialTransaction::write([1]),
    SerialTransaction::write([2]),
    SerialTransaction::flush(),
];
```

src/serial.rs
mciantyre: Sorry I wasn't clear. The goal is to maintain API consistency. Both a SPI and I2C mock each require a lifetime in the type signature, since they're based on the generic mock implementation. We add a lifetime here only to keep the APIs consistent.

The phantom marker is meaningless; we would compile fine without it. Should we consider breaking this consistency?

src/serial.rs
mciantyre: I'll add them.

Are we looking to guarantee the panic messages as part of the public API? Otherwise, what value will this provide? This might seem inconsistent with the SPI and I2C mock testing approaches.

src/serial.rs
mciantyre: Thanks, I hadn't considered re-using the mock. It's definitely valid and useful after giving it a second look! I'll fix-up `done()`.

mciantyre:
Thanks for your thorough review! Let me know what you think about round 2.

src/serial.rs
dbrgn: Yeah, I wouldn't include a lifetime parameter that isn't actually used / needed. In practice you don't explicitly pass in a lifetime anyways, so for the user the absence of the lifetime parameter doesn't break API consistency.

src/serial.rs
dbrgn: Yep, exactly :slightly_smiling_face: This is not about guaranteeing panic messages, it's about testing that the right panics happen.

ryankurte:
> @ryankurte do you know of a way to overcome the limitations listed above? If not, I'd suggest merging this PR once all open issues are resolved.

nothing immediately obvious, i've run into the same problem over in #23. my vote would be to merge and refactor later if someone has inspiration as to how ^_^

src/serial.rs
mciantyre: You got it. Removed in f426d65.

dbrgn:
Great, thank you @mciantyre!

dbrgn:
Version 0.6.0 released including this PR.

dbrgn:
Good timing @mciantyre, I just started writing a device driver using serial communication yesterday, and this PR proved to be very useful! :tada: :slightly_smiling_face:

PR #22, opened 2019-02-28
Friendlier expectations

ryankurte:
This PR alters the generic expectation module to accept `IntoIterator<Item=T>` types, allowing vector or slice references to be used with internal storage to avoid lifetime issues where expectations needed to outlive mock instances.


ryankurte:
This is slightly incompatible with #18, whichever is merged second will need to be updated to match.

PR #32, opened 2020-05-11
serial: Implement error expectations

dbrgn:
In contrast to the i2c/spi error expectations, the expected errors are embedded in the `Mode` enum. The reason for this is that serial transactions may consist of multiple reads/writes.

CC @mciantyre, maybe you want to take a look as well (as the original author of the serial mock).

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1,16 +1,14 @@
-//! Serial mock implementations
+//! Serial mock implementations.
 //!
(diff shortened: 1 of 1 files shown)


mciantyre:
LGTM! Thanks for the CC.

dbrgn:
Thanks for reviewing!

PR #33, opened 2020-06-18
Adapt to embedded-hal-1.0.0-alpha.1

eldruin:
Here the adaptions necessary for embedded-hal `1.0.0-alpha.1`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: rust:1.31
+      - image: rust:1.35
     steps:
       - checkout
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,5 @@ include = [
 edition = "2018"
 
 [dependencies]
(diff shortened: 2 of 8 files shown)


dbrgn:
Thanks! I think requiring Rust >=1.34 would be fine, what do you think?

eldruin:
Good point. The MSRV in `embedded-hal` had to be raised to Rust 1.35 due to an issue with associated constants in the ADC trait. So far that will be the MSRV of `embedded-hal` 1.0.0 so actually I would say the MSRV of this crate should be raised to 1.35 as well.
I will revert the commit adding void, update the MSRV and bump the minor version of this crate in this branch.

dbrgn:
Makes sense!

eldruin:
Done :)

ryankurte:
Looks good! Blocked on https://github.com/rust-embedded/embedded-hal/issues/177 (which I've mentioning everywhere useful so we have a list of pending PRs when we ship `v1.0.0`)

dbrgn:
@niclashoyer updated your branch to the latest alpha, so let's continue this discussion in #39.

PR #35, opened 2021-01-09
[WIP] Add timer mock

barafael:
I needed this for a project, switched to embedded-time now, but FWIW here is a very basic implementation of a timer mock


dbrgn:
Hi. Any reason for closing this? I didn't get around to review it yet, sorry 🙂

barafael:
Hi :) after making the changes I realized that Timer is not the right approach for me. So I won't be working on this for now. And I'm not sure the approach is right at all :D but feel free to have a look! It's not much.

dbrgn:
Ok, thanks for the info 🙂

PR #36, opened 2021-04-04
Fix some clippy lint complaints

barafael:


dbrgn:
Thanks for the changes. However, this includes a few CI features that do not yet work on Rust 1.31 (which is the currently supported Rust version), that's why the tests don't pass.

I'm open to conservative version bumps, but as long as it's easily possible, I like to support older Rust versions. Could you check what the minimal Rust version is that supports `cloned()` and `matches!`?

barafael:
Ok, I was unaware of that constraint.

`matches!` is available in stable since `1.42.0` as per the docs: [https://doc.rust-lang.org/std/macro.matches.html](https://doc.rust-lang.org/std/macro.matches.html), I confirmed with godbolt.

`cloned` on the other hand is officially available since 1.1.0, but apparently works on 1.0.0 too (again, godbolt).

dbrgn:
Could you bump the version in `README.md` and `.circleci/config.yml` to 1.42?

barafael:
I did it. Though I am not sure if replacing a `match` block with the `matches!` macro is worth such a version bump. OTOH, who would use 1.31 still?

eldruin:
Since it is just one instance and a couple simple lines, I would prefer keeping compatibility with Rust 1.31.0. That is what I use in almost all of my drivers at the moment and that would keep me using the last version of this crate.
I have no real need for Rust 1.31.0 other than that, so feel free to do whatever you want anyway.
For the record, when `embedded-hal` 1.0 comes out, the MSRV will be bumped at least to 1.35.0.

barafael:
I agree. I reverted the version bump and match change so it works with 1.31 again.

barafael:
I've squashed the changes into one commit; let me know if there is anything missing so this PR can be closed :)

dbrgn:
Thank you!

PR #39, opened 2021-12-21
Adapt for embedded-hal 1.0.0-alpha.6

niclashoyer:
Based on #33, tests are passing, but I don't have tested any further

Blocked on https://github.com/rust-embedded/embedded-hal/issues/177.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: rust:1.31
+      - image: rust:1.35
     steps:
       - checkout
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: rust:1.35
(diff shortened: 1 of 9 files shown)


dbrgn:
Thanks for the updates. MSRV may need to be updated again.

I closed #33 in favor of this PR, but of course, we'll wait with merging until 1.0.0 (or an RC) is released.

ryankurte:
hey @dbrgn what do you think of running an `alpha` branch here to match `e-h` pre `1.0.0`? we could merge this there which would make using `e-h-m` a bunch easier

dbrgn:
@ryankurte sorry for the late reply. Yes, I think that makes sense! I'll prepare such a branch.

dbrgn:
Check out https://github.com/dbrgn/embedded-hal-mock/tree/1-alpha! I also added tags starting with `eh-1.0.0-alpha`: https://github.com/dbrgn/embedded-hal-mock/tags

@niclashoyer @eldruin your commits are now part of the branch. PRs with updates to the latest alpha can be created against that branch.

PR #40, opened 2022-01-18
Add mock for timers

niclashoyer:
Added a mock for timers using [embedded-time](https://github.com/FluenTech/embedded-time/) with nanosecond precision. A single clock can be used to create multiple timers. The clock can be freely skipped by any time value supported by embedded_time (e.g. microseconds, seconds, nanoseconds, …).

To support sharing a clock and timer between threads I relied on `Arc<AtomicU64>` for the time base. That also allows each timer to actually own the clock, so that they can be freely shared with any driver implementation.

I added this on top of #39. If wanted, I could also try to backport this.


niclashoyer:
I think embedded_time needs a newer rust version. They don't have a MSRV given, so we need to figure that out somehow.

dbrgn:
Sorry for the long delay. I'm quite busy at the moment (as always 😂), so if someone else with timer experience would do a first review, that would already be helpful. (Anyone can do a review, not just maintainers 🙂)

The branch also needs a rebase.

eldruin:
In the mean time we have decided to [remove timers in `embedded-hal` 1.0](https://github.com/rust-embedded/embedded-hal/issues/359) so I think it would not make much sense to merge this as-is. 
Something that would make sense, though, would be to adapt the mock proposed here to the version available in `embedded-hal` 0.2.7. Pretty much cherry picking commit [b8e03b8](https://github.com/dbrgn/embedded-hal-mock/pull/40/commits/b8e03b854cbea2f6273ad14e41a25fce6a652e9a) on top of master would work as the interface was the same.
From a quick look the mock proposed here seemed fine but I can do a proper review of this afterwards.
One additional thing to think about would be whether to use `embedded-time` or [`fugit`](https://github.com/korken89/fugit) (or both).

dbrgn:
@niclashoyer would you be interested in doing the backport suggested by @eldruin?

niclashoyer:
@dbrgn @eldruin done! I'm currently moving to `embedded-hal-async`, but I think it still makes sense to backport this.

dbrgn:
Sorry for the long delay again.

I rebased your branch against current master and will now review the contribution 🙂

Regarding the added dependencies, I'm thinking about putting the timer feature behind a cargo flag.

dbrgn:
I pushed two changes:

- For consistency with other modules, I've renamed `SimClock` to `MockClock` and `SimTimer` to `MockTimer`.
- I put the module behind a Cargo feature (enabled by default)

Thanks! 👍🏻

PR #42, opened 2022-05-05
spi: Adds mock implementation for WriteIter

silvergasp:
Happy for a few runs at this as I'm relatively new to rust.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -193,11 +193,35 @@ impl spi::Transfer<u8> for Mock {
     }
 }
 
+impl spi::WriteIter<u8> for Mock {
+    type Error = MockError;
+
+    fn write_iter<WI>(&mut self, words: WI) -> Result<(), Self::Error>
+    where
+        WI: IntoIterator<Item = u8>,
+    {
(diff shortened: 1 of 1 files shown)


dbrgn (changes requested):
Thanks! I wonder if you could delegate to the regular `Write` implementation, similar to the way it was done with I2C here: https://github.com/dbrgn/embedded-hal-mock/pull/44

dbrgn:
@silvergasp are you interested in doing the suggested change, or should someone else take over your branch?

silvergasp:
Oh yeah sorry I missed this one, I'll take another pass at it

dbrgn:
@silvergasp friendly reminder 🙂 If you don't have time, just let me know.

dbrgn:
Hmm, I just pushed a version that calls self.write, but I realized that while it re-uses code, it makes error reporting worse (since error messages on missing expectation will mention `spi::write`, and not `spi::write_iter`). I'll undo that commit.

dbrgn (approved):
Thanks, I think this can now be merged!

PR #46, opened 2022-08-07
spi: add Transaction::read_vec

newAM:
At the moment SPI reads with a length >1 cannot be represented.


dbrgn (approved):
Thanks! I guess this was an oversight 🙂

PR #47, opened 2022-09-19
nb: use embedded-hal re-export

newAM:
The `embedded-hal` alpha added a re-export for `nb`.  Using the export will ensure that `nb` versions are always the same between this crate and `embedded-hal`.


dbrgn (approved):
Thanks!

PR #50, opened 2022-10-31
Replace circleci with github actions

newAM:
For #49, as far as I know github actions has to be used on the default branch before it will be available on other branches.

This tests with nightly in addition to stable - which isn't necessary for the `master` branch, but it doesn't hurt, and it keeps the CI files aligned.


.github/actions/ci.yml
```diff
@@ -0,0 +1,61 @@
+on:
+  # trigger on pushes to the master branch and 1-alpha
+  push:
+    branches:
+      - main
+      - 1-alpha
+  # trigger on all pull requests
+  pull_request:
+  # enable manual triggering
+  workflow_dispatch:
```
dbrgn, line 10: Interesting, TIL!

dbrgn (changes requested):
Thank you, looks good! Can you update the README as well? The badges still reference CircleCI.

newAM:
> Thank you, looks good! Can you update the README as well? The badges still reference CircleCI.

Good catch, done!

dbrgn:
Looks good! Let's merge this, to see if it works 🙂

PR #52, opened 2022-12-04
add `PwmPin` mock

rursprung:
this extends the `Pin::Mock` to also cover the [`PwmPin`][] trait.

some observations on this commit:
* i took the liberty of just hardcoding the `PwmPin::Duty` to be an
  `u16` (wrapped in the `PwmDuty` type because it's used in other
  places here in the module as well). i don't see an easy way to make
  this configurable because i'd have to add it as a generic attribute to
  `Pin` and then it'd have to be defined by everyone using it
  (unnecessary for all non-PWM use-cases). but i think this should be
  acceptable as u16 probably covers most/all use-cases.
* the current code features quite some code duplication (essentially the
  method check implementations are all the same for all setters / all
  getters). i've continued this for now but it might be worth a
  refactoring in the future (i haven't touched it because i'm not sure
  if there's some strategic decision behind this?)
* there's a `TransactionKind::is_get` API for which i don't really see
  the reason (why not just check it directly?). i've decided to not copy
  that approach for the others as it'd IMHO just bloat the code and i'd
  instead suggest to remove `TransactionKind::is_get` at a later point.
* `PwmPin` only exists on `embedded-hal` 0.x, it is currently missing
  from the planned 1.0 release. see rust-embedded/embedded-hal#358 for
  the discussion on this. once it has (hopefully) been re-added there a
  corresponding mock can be provided here also for 1.x.

[`PwmPin`]: https://docs.rs/embedded-hal/0.2.7/embedded_hal/trait.PwmPin.html

### DIFF
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -43,6 +43,10 @@ use crate::common::Generic;
 use crate::error::MockError;
 
 use embedded_hal::digital::v2::{InputPin, OutputPin};
+use embedded_hal::PwmPin;
+
+/// The type used for the duty of the [`PwmPin`] mock.
+pub type PwmDuty = u16;
 
 /// MockPin transaction
 #[derive(PartialEq, Clone, Debug)]
@@ -80,10 +84,42 @@ impl Transaction {
         Transaction::new(TransactionKind::Set(state))
     }
 
+    /// Create a new disable transaction
+    pub fn disable() -> Transaction {
+        Transaction::new(TransactionKind::Disable)
+    }
+
+    /// Create a new enable transaction
+    pub fn enable() -> Transaction {
+        Transaction::new(TransactionKind::Enable)
+    }
+
+    /// Create a new get_duty transaction
+    pub fn get_duty(duty: PwmDuty) -> Transaction {
+        Transaction::new(TransactionKind::GetDuty(duty))
+    }
+
+    /// Create a new get_max_duty transaction
+    pub fn get_max_duty(max_duty: PwmDuty) -> Transaction {
+        Transaction::new(TransactionKind::GetMaxDuty(max_duty))
+    }
+
+    /// Create a new set_duty transaction
+    pub fn set_duty(expected_duty: PwmDuty) -> Transaction {
+        Transaction::new(TransactionKind::SetDuty(expected_duty))
+    }
+
     /// Add an error return to a transaction
     ///
     /// This is used to mock failure behaviours.
+    ///
+    /// Note that this can only be used for methods which actually return a [`Result`];
+    /// trying to invoke this for others will lead to an assertion error!
     pub fn with_error(mut self, error: MockError) -> Self {
+        assert!(
+            self.kind.supports_errors(),
+            "the transaction kind supports errors"
+        );
         self.err = Some(error);
         self
     }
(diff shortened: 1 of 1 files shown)


dbrgn:
The `PwmPin` trait doesn't seem to exist anymore in e-h 1.0. Do you know what the replacement is, if any?

(Note: I'm happy to merge PRs for 0.2 as well, but that mock will become obsolete once e-h 1.0 is out.)

rursprung:
> The `PwmPin` trait doesn't seem to exist anymore in e-h 1.0. Do you know what the replacement is, if any?

hm, i wasn't aware of that (never looked into e-h 1.0 so far). i'll have to dig into that (probably in the next few days).
i also just saw that you're testing against rust 1.31 and that doesn't contain `u16::MAX` yet (i just used the latest rust stable and didn't bother checking as i didn't expect to be using anything new). should i just use a hard-coded value there or do you want to lift the MSRV instead (which is a bit stupid because i'm only using it in a test and not in the runtime code :/)

dbrgn:
I think `main` is the "legacy branch", and if it's only for testing, I'd use a constant instead for now.

rursprung:
@dbrgn: i've updated the PR a while ago so that it now should build fine with rust 1.31. i've also pushed the e-h 1.0 ticket but so far no updates there. would it be ok for you to take this into 0.x for the time being? based on the e-h discussions it seems that some form of `PwmPin` should eventually land in e-h 1.x again (though i do hope that the discussion leads to it already landing in 1.0).

rursprung:
@dbrgn any chance that you could already review this for e-h 0.x while the e-h 1.x discussion is ongoing? as e-h 1.x will anyway look differently it'll require a different implementation here (if i don't forget i'll probably provide it here as well once a solution lands in e-h 1.x).
i'd like to release my motor driver (which would need this) but don't want to have a git-dependency in it (even if it's just for the test).
thanks!

dbrgn:
Sorry, I missed your original comment. Thanks for the reminder! I'll take a look tonight.

src/pin.rs
```diff
@@ -187,15 +244,73 @@ impl InputPin for Mock {
...
+        let mut s = self.clone();
+
+        let Transaction { kind, err: _err } =
+            s.next().expect("no expectation for pin::get_max_duty call");
+
+        if let TransactionKind::GetMaxDuty(max_duty) = kind {
+            max_duty
+        } else {
+            assert!(false, "expected pin::get_max_duty");
+            unreachable!()
```
dbrgn, line 288: Instead of `assert!(false, msg)` followed by `unreachable!()`, simply use `panic!(msg)`. It's the same.

```suggestion
            panic!("expected pin::get_max_duty");
```

Same thing counts for other methods as well.

src/pin.rs
```diff
@@ -235,4 +350,23 @@ mod test {
...
+            Transaction::new(GetDuty(expected_duty)),
+            Transaction::new(Disable),
+        ];
+        let mut pin = Mock::new(&expectations);
+
+        pin.enable();
+        let max_duty = pin.get_max_duty();
+        pin.set_duty(max_duty);
+        assert_eq!(pin.get_duty(), expected_duty);
+        pin.disable();
```
dbrgn, line 370: You missed the call to `pin.done()` at the end 🙂

dbrgn (changes requested):
I left some comments. Once these things are fixed, I can tag a new release soon!

rursprung (commented):
thanks for your review! i've fixed all your findings

src/pin.rs
dbrgn: Yeah, Rust's assertions are not too fancy unfortunately 🙂

dbrgn (approved):
👍🏻 thanks for the updates!

dbrgn:
> implicitly i'm also asking for a release of this crate after this PR gets merged angel

@rursprung 0.9.0 is released!

PR #56, opened 2023-01-07
enable GitHub Dependabot

rursprung:
this ensures that the dependencies are kept up to date. see [the docs][] for further information.

note: there's intentionally no entry in the changelog as it does not have an impact on consumers.

[the docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"


rursprung:
this should then e.g. also take care of upgrading `nb` to 1.x (note that `nb v0.1.3` is already a wrapper around `nb v1.0.0`, so this would then just ensure that this library here depends on a stable release).

dbrgn:
Regarding `nb`: Since `embedded-hal` 0.2.7 depends on `nb` 0.1.3, I assume we shouldn't depend on a higher version than that, right?

If `0.1.3` and `1.x` can be used interchangeably, then `embedded-hal` should bump the version first.

rursprung:
> Regarding `nb`: Since `embedded-hal` 0.2.7 depends on `nb` 0.1.3, I assume we shouldn't depend on a higher version than that, right?

i don't see an issue here as e-h 2.2.7 explicitly depends on nb 0.1.3 which in turn is using nb 1.0.0. i just checked, and e-h-m compiles just fine either way.

however, this shouldn't impact the dependabot discussion per-se as dependabot just raises PRs, it can then still be told to ignore a certain version (e.g. major version) of a specific dependency if you don't want that upgrade for the time being.

dbrgn:
Yeah, let's merge this for now 🙂 I think when Dependabot tries to upgrade `nb`, I'd stick to the version used by `e-h`.

PR #57, opened 2023-01-07
define the MSRV in `Cargo.toml`

rursprung:
this way the compiler will issue a clear error message to consumers using outdated rust versions (rather than cryptic compile errors due to missing features).

also slightly adapted the README to more closely match the MSRV definition of other big crates there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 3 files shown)


clippy.toml
```diff
@@ -1 +0,0 @@
-msrv = "1.60"
```
dbrgn, line 1: Don't remove this! The goal of this field is that even if you run Clippy on Rust 1.80, Clippy will not suggest things that won't compile on 1.60. (Otherwise, depending on your local Rust version, you'll get different Clippy warnings.)

dbrgn (changes requested):

clippy.toml
rursprung: ah, i didn't realize that - thanks! i'll patch it up tomorrow and re-push it then

dbrgn (approved):
Rebased your commit and properly formatted the CHANGELOG.

Thanks!

PR #59, opened 2023-01-07
Fail if a mock is dropped without having called `.done()` exactly once

### LINKED ISSUE #34: Implement `Drop` for mocks which executes `done()`
When writing tests it's easy to forget a call to `done()` at the end to verify if there are any expectations left. By executing it in drop one can't forget it.

```rust
// will pass if no transaction happens 
#[test]
fn test() {
        let expectations = [
            Transaction::write(0x58, vec![0xab, 0xcd]),
        ];
        let mut mock = I2cMock::new(&expectations);
}
```

dbrgn:
Not calling `.done()` is a potential bug. Print a warning to stderr and fail the test if `done()` is not called exactly once.

I first thought about putting the strict checking behind a feature flag, but it's better to simply release a breaking release (0.10) instead.

Fixes #34.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,12 +4,13 @@ use std::{
     collections::VecDeque,
     fmt::Debug,
     sync::{Arc, Mutex},
+    thread,
 };
 
 /// Generic mock implementation.
 ///
-/// ⚠️ **Do not use this directly as end user! This is only a building
-/// block for creating mocks.**
+/// ⚠️ **Do not use this directly as end user! This is only a building block
+/// for creating mocks.**
 ///
 /// This type supports the specification and evaluation of expectations to
 /// allow automated testing of hal drivers. Mismatches between expectations
@@ -22,6 +23,7 @@ use std::{
 #[derive(Debug)]
 pub struct Generic<T: Clone + Debug + PartialEq> {
     expected: Arc<Mutex<VecDeque<T>>>,
+    done_called: Arc<Mutex<DoneCallDetector>>,
 }
 
 impl<'a, T: 'a> Generic<T>
@@ -37,6 +39,7 @@ where
     {
         let mut g = Generic {
             expected: Arc::new(Mutex::new(VecDeque::new())),
+            done_called: Arc::new(Mutex::new(DoneCallDetector::new())),
         };
 
         g.expect(expected);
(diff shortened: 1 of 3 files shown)


src/common.rs
```diff
@@ -87,19 +96,93 @@ where
...
+            use std::io::Write;
+            let mut stderr = std::io::stderr();
+            stderr.write_all(b"\x1b[31m").ok();
+            stderr.write_all(msg.as_bytes()).ok();
+            stderr.write_all(b"\x1b[m\n").ok();
+            stderr.flush().ok();
+
+            // Panic. This probably results in an abort:
+            // https://doc.rust-lang.org/std/ops/trait.Drop.html#panics
+            panic!("{}", msg);
```
rnestler, line 142: Why should this result in an abort? Only if we *not* check for `!thread::panicking`, right?

src/common.rs
```diff
@@ -4,12 +4,13 @@ use std::{
...
     sync::{Arc, Mutex},
+    thread,
 };
 
 /// Generic mock implementation.
 ///
-/// ⚠️ **Do not use this directly as end user! This is only a building
-/// block for creating mocks.**
+/// ⚠️ **Do not use this directly as end user! This is only a building block
+/// for creating mocks.**
```
rnestler, line 13: Why is it `pub` if it isn't supposed to be used as and user?

dbrgn: Because our mocks are aliases too concrete instances of `GenericMock<T>`, for example `pub type Mock = Generic<Transaction>`.

If we change the visibility to `pub(crate)` and try to compile:

```
error[E0446]: crate-private type `common::Generic<eh0::spi::Transaction>` in public interface
   --> src/eh0/spi.rs:120:1
    |
120 | pub type Mock = Generic<Transaction>;
    | ^^^^^^^^^^^^^ can't leak crate-private type
    |
   ::: src/common.rs:24:1
    |
24  | pub(crate) struct Generic<T: Clone + Debug + PartialEq> {
    | ------------------------------------------------------- `common::Generic<eh0::spi::Transaction>` declared as crate-private
```

src/common.rs
dbrgn: Yeah, I think you're right.

dbrgn: Note updated in https://github.com/dbrgn/embedded-hal-mock/pull/79

PR #63, opened 2023-01-07
Generic mock: When calling expect, call done internally

dbrgn:
When updating expectations on a mock by calling `.expect(...)` on it, assert that previous expectations have been consumed. (We discussed this in Matrix, @ryankurte.)

Open questions:

- Should we rename `expect`, to avoid confusion with the method with the same name on `Result` and `Option`? I'd suggest `update_expectations`. This is a breaking change, but it's a semver-breaking release anyways.
- Should it be legal to call  `.expect(...)` after calling `.done(...)`? For now, I added a check which ignores double-done when calling `.expect(...)`.

Another issue: Right now we have two separate state variables in an `Arc<Mutex<_>>` that need to be locked independently. This smells racy, it would probably be good to have a single lock for the internal shared state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+
 - Implement mock for `embedded_hal::pwm::SetDutyCycle`
 
 ### Fixed
 
 ### Changed
+
+- When updating expectations on a mock by calling `.expect(...)` on it, assert
+  that previous expectations have been consumed (#63)
 
 
 ## 0.10.0-rc.1 - 2023-11-01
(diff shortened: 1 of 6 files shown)


ryankurte:
looks generally good, what is the advantage of a `done` flag as opposed to checking expectations match then just clearing out the expectation buffer?

> Should we rename expect, to avoid confusion with the method with the same name on Result and Option?

i'm okay with the collision or with renaming this as proposed, that said, `expect` was the most accurate short name i could come up with  and `update_expectations` seems unweildy in comparison?

> Should it be legal to call .expect(...) after calling .done(...)?

i think so? if you break down a test into steps it might be neat to separate tests by `done`, though it does not make any difference to the execution.

dbrgn:
It's been a while, sorry for the delay 🙂

> looks generally good, what is the advantage of a done flag as opposed to checking expectations match then just clearing out the expectation buffer?

Hmm, I don't quite understand what you mean... Could you elaborate?

> i'm okay with the collision or with renaming this as proposed, that said, `expect` was the most accurate short name i could come up with and `update_expectations` seems unweildy in comparison?

I don't find it much worse than `expect`, and I could imagine that the naming conflict with Result/Option could trip up a few people 🙂 I updated the MR with a rename commit.

PR #65, opened 2023-01-10
Fix documentation url in 'Cargo.toml' file

regexident:
The documentation URL currently points to the `embedded-hal` crate, rather than `embedded-hal-mock`


dbrgn:
Haha, nice catch 😄 Thanks!

PR #66, opened 2023-04-15
update to `embedded-hal:1.0.0-alpha.10`

rursprung:
note that i updated this purely based on the build errors and the [release notes].

[release notes]: https://github.com/rust-embedded/embedded-hal/releases/tag/v1.0.0-alpha.10

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,5 @@ include = [
 edition = "2018"
 
 [dependencies]
-embedded-hal = "=1.0.0-alpha.9"
-embedded-hal-nb = "=1.0.0-alpha.1"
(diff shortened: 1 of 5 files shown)


rursprung:
@dirbaio and @eldruin: nearly all commits in e-h 1.0.0-alpha.10 come from you. could you please double-check if the changes here match the expectations you have for the implementation of these APIs? thanks!

this is a preparation to later add the mock implementation of the new e-h 1.0 PWM APIs (which are new with this release, so the update is a pre-requisite).

newAM (approved):
Was just about to do this myself, glad I checked for open PRs first!

Looks good to me!

dbrgn:
Great, thanks @rursprung @newAM!

Issue #13, opened 2018-12-26
Rename delay::MockNoop to delay::NoopDelay


dbrgn:
Rename it to `NoopDelay`.

Add new alias and deprecate the old one in `0.5`, then remove it in `0.6`.

PR #67, opened 2023-05-09
Rename delay::MockNoop to delay::NoopDelay #13

tomgilligan:


src/delay.rs
```diff
@@ -18,24 +18,26 @@ use std::time::Duration;
 use embedded_hal::blocking::delay;
 
 /// A `Delay` implementation that does not actually block.
-pub struct MockNoop;
+pub struct NoopDelay;
+/// Deprecated name for no-op `Delay` implementation.
+pub type MockNoop = NoopDelay;
```
dbrgn, line 23: Hmm, I just realized that a type alias can't be used as constructor. Thus, `let m = MockNoop;` will not compile.

Instead, let's just remove `MockNoop`, breaking changes are OK. However, please add a note to `CHANGELOG.md` about the rename.

dbrgn (changes requested):
Thanks! Left a note.

PR #69, opened 2023-05-16
Stricter panic assertions

tomgilligan:
Assertions on the message of the panic could be helpful for further development eg. adding labels to expectations (they would be included in messages)
Also helps tests document themselves better without comments.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -161,24 +161,25 @@ mod tests {
         }
 
         #[test]
-        #[should_panic]
+        #[should_panic(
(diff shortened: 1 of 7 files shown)


dbrgn:
Thanks, this looks like a good idea!

Could you rebase your branch against current `main`?

dbrgn:
> This feels like maybe it's a bit extra. Please let me know your preference.

I'm a bit unsure. I probably would have included only substrings, but the full messages also have some advantages, so I'll merge this as-is for now.

PR #74, opened 2023-07-02
add `SetDutyCycle` mock

### LINKED ISSUE #73: add mock for e-h 1 `SetDutyCycle` (`PwmPin` successor) trait
https://github.com/rust-embedded/embedded-hal/pull/430 introduced the `SetDutyCycle` trait as an embedded-hal 1.0 replacement for the previous 0.2 `PwmPin` (for which a mock implementation exists).
it'd be great to have a mock implementation also for the new trait!

note: #66 already updated this repo to the required e-h 1.0 pre-release.

rursprung:
this is the 1.0 equivalent of the old `PwmPin` trait (the mock for this has been added in #52).

note that the test coverage is fully handled by the doc test which is why there's no additional `mod test` in this file.

this fixes #73

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Implement mock for `embedded_hal::pwm::SetDutyCycle`
 
 ## 0.8.0 - 2021-08-16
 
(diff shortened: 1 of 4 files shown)


rursprung:
> I feel like it makes more sense for expectations to only be consumed by a matching method. Having said that, I think what you have here is consistent with other parts of `embedded-hal-mock`. Just wanted to highlight this as something to potentially discuss outside of this PR.

i contemplated adding mocks for _all_ methods of the trait - but then that wouldn't reflect how the trait is usually implemented: normally the HALs will probably not overwrite the default implementations.
i could've probably added a test which used the methods directly, but this way it's also visible that it's being used by the other methods underneath.

then again, this sort-of exposes an implementation detail of embedded-hal... so maybe i just picked the lazy option? 😅🤷

dbrgn:
Hi, would you mind targeting `main` instead, now that #75 was merged?

rursprung:
superseded by #88 which now targets the `main` branch

Issue #70, opened 2023-05-19
Query: EH1 alpha release


tommy-gilligan:
Thank you for having a EH1 alpha branch that is up-to-date with the latest EH1 alpha.  It has been very useful in some work I've been doing.
I think it would be nice to release this work soon (perhaps a version of embedded-hal-mock that supports both EH0.2 and EH1.0-alpha).  I'm happy to sync the branch with main if that is at all helpful but also totally understand if a merge is not planned for anytime soon.  I would love to know your thoughts on this.

newAM:
I had the same thought (merging eh0.2 and eh1a into the same branch).  Even after eh1 is released 0.2 will be around a long time and it would be nice to support both in this crate.

I'm also happy to help out with merge work to make this happen.

ryankurte:
yeah supporting both is an interesting question... it's always in need of updates but, is this something we can support just by keeping `e-h-m` against alpha then either recommending or pulling in https://github.com/ryankurte/embedded-hal-compat ?

newAM:
I envision this being two separate implementations for each HAL.  For exmaple, I have an eh0 and an eh1 module in my W5500 crate: https://github.com/newAM/w5500-rs/tree/main/ll/src

rursprung:
having a release of the mock implementation for e-h 1.0.0-alpha.10 would be very useful as i'd like to add support for it to my crates but i need the mock implementations for CI.

is there any chance to create a corresponding alpha release off of the alpha branch for now?

IMHO going forward merging the two together would be easiest, as is also done by the HAL implementations (e.g. [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal) is doing it like this)

PR #75, opened 2023-07-05
Support embedded-hal 0.x and 1.0-alpha.x

tomgilligan:
Rebased dbrgn 1-alpha branch

This moves support for different versions of embedded-hal to separate modules `eh0` and `eh1` as suggested at https://github.com/dbrgn/embedded-hal-mock/issues/70#issuecomment-1555003903


rursprung:
see my comment on https://github.com/dbrgn/embedded-hal-mock/issues/70#issuecomment-1616625227 for why i (now) think that _not_ merging e-h 0 and e-h 1 mocks in the same e-h-m release would be better.

rursprung:
fair enough. though then it might be worth it to move eh0 stuff into the eh0 module and either have all eh1 things top-level or at least re-export them at the top-level so that they don't need a breaking change in the future (and this way e-h-m can match e-h in terms of API structure). WDYT?

newAM (approved):
This looks great!

src/eh1/serial.rs
```diff
@@ -0,0 +1,636 @@
...
+//! serial traits. You can use the same mock for both interfaces.
+//!
+//! ## Usage: Non-blocking serial traits
+//!
+//! ```
+//! # use eh1 as embedded_hal;
+//! // Note that we're using the non-blocking serial traits
+//! use embedded_hal_nb::serial::{Read, Write};
+//! use embedded_hal_mock::eh1::serial::{
+//!     Mock as SerialMock,
```
ryankurte, line 16: this might be how we've always done it but, it seems like it's worth having descriptive names at the module level to avoid every consumer having to `Mock as SerialMock` ?

dbrgn: I'd rather keep PRs focussed on one thing, makes it easier to review 🙂

The original idea was to implement the same pattern for every mock. However, I guess you're right and if every user aliases the import, we can rename the mocks as well. So: A separate PR would be welcome (but it might make sense to base it on this branch, or to wait until it is merged.)

dbrgn:
Thanks for your work, @tomgilligan and others! This looks great at first glance! I'd like to take a closer look before merging (since it's a major change), but don't know yet when I find time to do that. Hopefully by next week... If I forget about it, feel free to ping me in this thread at any time! 🙂 (I do get notifications, but sometimes they drown in the flood of e-mails.)

dbrgn:
I did take a closer look now, and also ported one of my project's tests to your branch. It is looking good!

Let's merge this branch, and I'll try to publish a release (maybe an `rc1`) soon. Thanks for your work!

dbrgn (approved):

dbrgn:
This is now released as `embedded-hal-mock` version `0.10.0-rc.1`: https://crates.io/crates/embedded-hal-mock/0.10.0-rc.1

PR #76, opened 2023-07-05
update to `embedded-hal=1.0.0-alpha.11`

rursprung:
see [their changelog][eh-changelog] for further details.

[eh-changelog]: https://github.com/rust-embedded/embedded-hal/blob/v1.0.0-alpha.11/embedded-hal/CHANGELOG.md

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -11,7 +11,7 @@
 //! extern crate embedded_hal;
 //! extern crate embedded_hal_mock;
 //!
-//! use embedded_hal::spi::{SpiBus, SpiBusWrite};
+//! use embedded_hal::spi::SpiBus;
 //! use embedded_hal_mock::spi::{Mock as SpiMock, Transaction as SpiTransaction};
 //! use embedded_hal_nb::spi::FullDuplex;
 //!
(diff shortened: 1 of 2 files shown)


src/spi.rs
```diff
@@ -408,16 +359,19 @@ impl spi::SpiDevice for Mock {
...
                 Operation::Transfer(read, write) => {
-                    spi::SpiBus::transfer(self, read, write)?;
+                    SpiBus::transfer(self, read, write)?;
                 }
                 Operation::TransferInPlace(buffer) => {
-                    spi::SpiBus::transfer_in_place(self, buffer)?;
+                    SpiBus::transfer_in_place(self, buffer)?;
+                }
+                Operation::DelayUs(_) => {
+                    todo!();
```
rursprung, line 374: @Dirbaio: you introduced this new operation in https://github.com/rust-embedded/embedded-hal/commit/d07d39e359ed2157cbc0bbd88ef950b60c58e804
could you please advise on how this is best covered here? (sorry, i'm not at all familiar with the SPI implementation - i just want to bump this to the newest e-h version before the release of e-h-m for 1.0.0-alpha)

thanks!

Dirbaio: In non-mock code you'd either make it generic on a `T: DelayUs`, or if the implementation is hardware-specific directly use a hardware-specific delay. If for whatever reason you can't, returning a "delay not supported" error is also an option.

In the mock case you might want to check it against the expectations? so the user can assert the delay is done. Not sure.

eldruin: For the case of making it generic over `T: DelayUs`, this crate provides a no-op delay mock, which one could use for this.

rursprung:
thanks all for your feedback! i've now added a new transaction type `Delay` for the mock and also added the test coverage for this. this is now ready for review.

newAM (approved):
Works for me with my W5500 crate.

dbrgn (changes requested):
Thanks for the contribution!

I'm not too familiar with SPI, so this is only a superficial review. However, I noticed a few issues.

Reviews by other users or contributors are welcome! Otherwise I'd merge this PR once the open issues are resolved.

dbrgn:
Additionally, the changes from #75 were merged into `main`, so: Do you have time to update your branch?

newAM:
@rursprung let me know if you want help with this :)

newAM:
I opened #85 with the feedback applied & rebased onto `main`.  Feel free to use any part of that.

rursprung:
fully covered in #86 - thanks @newAM for picking this up while i was away!

PR #78, opened 2023-07-22
wip: Add CAN mock 

schphil:
This is a initial draft for adding a CAN mock. Any feedback would be greatly appreciate.

### DIFF
--- a/src/can.rs
+++ b/src/can.rs
@@ -1,0 +1,252 @@
+//! CAN mock implementations.
+//!
+//! ## Usage
+//!
+//! ```
+//! extern crate embedded_hal;
+//! extern crate embedded_hal_mock;
+//!
+//! use embedded_can::blocking::Can;
+//! use embedded_can::Frame;
+//! use embedded_can::{StandardId, ExtendedId};
+//! use embedded_hal_mock::can::{Mock as CanMock, Transaction as CanTransaction, MockFrame as CanMockFrame};
+//!
+//! // Create CAN frame
+//!
+//! let id: StandardId = StandardId::new(0x123).unwrap();
+//! let frame1 = CanMockFrame::new(id, &[0x00, 0x00, 0x00, 0x00, 0x10]).unwrap();
+//! let id: ExtendedId = ExtendedId::new(0x12345678).unwrap();
+//! let frame2 = CanMockFrame::new(id, &[0x00, 0x00, 0x00, 0x00, 0x10]).unwrap();
+//!
+//! // Configure expectations
+//! let expectations = [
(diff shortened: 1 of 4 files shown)


src/can.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// An optional error return for a transaction.
+    ///
+    /// This is in addition to the mode to allow validation that the
+    /// transaction mode is correct prior to returning the error.
+    expected_err: Option<ErrorKind>,
+}
+
+impl Transaction {
+    /// Create a Transmit transaction
+    pub fn transmit(expected_frame: MockFrame) -> Transaction {
```
ryankurte, line 72: i think it probably makes sense to `const`ify everywhere possible to make it easier to construct test vectors?

src/can.rs
```diff
@@ -0,0 +1,272 @@
...
+        assert_eq!(
+            t.expected_mode,
+            Mode::Transmit,
+            "can::transmit unexpected mode"
+        );
+        assert_eq!(
+            &t.expected_frame.unwrap(),
+            frame,
+            "can::transmit data does not match expectation"
+        );
```
ryankurte, line 162: this looks pretty reasonable to me, tho i think is missing a path to return errors? (if `expected_error` is set this value should be returned instead of `Ok(())` (and the same for receive and the nb impls)

src/error.rs
```diff
@@ -15,6 +15,12 @@ impl embedded_hal::digital::Error for MockError {
     }
 }
 
+impl embedded_can::Error for MockError {
```
ryankurte, line 18: i suspect `MockError` is going to need to reflect the variants in [embedded_can::ErrorKind](https://docs.rs/embedded-can/latest/embedded_can/enum.ErrorKind.html) so mocks can deal with at least contention (`Bit`) and `CRC` errors... but we can probably expand on this later.

src/can.rs
```diff
@@ -0,0 +1,272 @@
...
+    Receive,
+}
+
+/// CAN Transaction type
+///
+/// Models an CAN transmit or receive transaction
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Transaction {
+    expected_mode: Mode,
+    expected_frame: Option<MockFrame>,
```
ryankurte, line 61: because it's either rx or tx there's only one frame (unlike SPI where you have both incoming and outgoing data) so it seems like `expected_frame` and `response_frame` could just be `frame`?

ryankurte (commented):
thanks for opening this! a few bits of feedback but overall seems good

PR #82, opened 2023-08-01
docs: add doc_auto_cfg

newAM:
This adds features hints in the documentation that gets published to docsrs.  e.g. [syn](https://docs.rs/syn/latest/syn/#modules)

Test locally with: `cargo rustdoc --all-features -- --cfg docsrs`


dbrgn:
Nice, thanks!

PR #83, opened 2023-08-01
README: update min-rust-badge

newAM:
Updates the min-rust-badge to match documentation and the `rust-version` field in Cargo.toml


dbrgn:
Good catch!

PR #84, opened 2023-08-01
README: remove dead links

newAM:
Alternatively I could link these up to docs.rs pages in anticipation of the next release.


dbrgn:
I think removing them makes sense for now. Thanks!

PR #88, opened 2023-09-13
add `SetDutyCycle` mock

### LINKED ISSUE #73: add mock for e-h 1 `SetDutyCycle` (`PwmPin` successor) trait

rursprung:
this is the 1.0 equivalent of the old `PwmPin` trait (the mock for this has been added in #52).

note that the test coverage is fully handled by the doc test which is why there's no additional `mod test` in this file.

this fixes #73

(note: this replaces #74 which was targeting the now defunct `1-alpha` branch)

### DIFF
--- a/src/eh1/pwm.rs
+++ b/src/eh1/pwm.rs
@@ -1,0 +1,127 @@
+//! Mock implementations for
+//! [`embedded_hal::pwm`](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/pwm/index.html).
+//!
+//! Usage example:
+//! ```
+//! use std::io::ErrorKind;
+//!
+//! # use eh1 as embedded_hal;
+//! use embedded_hal::pwm::SetDutyCycle;
+//! use embedded_hal_mock::eh1::{
+//!     pwm::{Mock as PwmMock, Transaction as PwmTransaction},
+//!     MockError,
+//! };
+//!
+//! // Configure expectations
+//! let expectations = [
+//!     PwmTransaction::get_max_duty_cycle(100),
+//!     PwmTransaction::set_duty_cycle(50),
+//!     PwmTransaction::set_duty_cycle(101).with_error(MockError::Io(ErrorKind::NotConnected)),
+//! ];
+//!
+//! // Create pin
+//! let mut pwm = PwmMock::new(&expectations);
+//!
+//! // Run and test
+//! pwm.set_duty_cycle_percent(50).unwrap();
+//! pwm.set_duty_cycle(101).expect_err("expected error return");
+//!
+//! // Finalise expectations
+//! pwm.done();
+//! ```
+
+use eh1::pwm::{ErrorKind, ErrorType, SetDutyCycle};
+
+use crate::{common::Generic, eh1::MockError};
+
+/// MockPwm transaction
+#[derive(PartialEq, Clone, Debug)]
+pub struct Transaction {
+    /// Kind is the transaction kind (and data) expected
+    kind: TransactionKind,
+    /// Err is an optional error return for a transaction.
+    /// This is in addition to kind to allow validation that the transaction kind
+    /// is correct prior to returning the error.
+    err: Option<MockError>,
+}
+
+impl Transaction {
+    /// Create a new PWM transaction
+    pub fn new(kind: TransactionKind) -> Transaction {
+        Transaction { kind, err: None }
+    }
+
+    /// Create a new [`TransactionKind::GetMaxDutyCycle`] transaction for [`SetDutyCycle::get_max_duty_cycle`].
+    pub fn get_max_duty_cycle(duty: u16) -> Transaction {
+        Transaction::new(TransactionKind::GetMaxDutyCycle(duty))
+    }
+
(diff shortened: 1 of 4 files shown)


rursprung:
it seems that the format of the assertion message is now different in nightly? i.e. the failure seems to be unrelated...
```
       panic message: `"assertion `left == right` failed: i2c::write unexpected mode\n  left: Read\n right: Write"`,
 expected substring: `"assertion failed: `(left == right)`\n  left: `Read`,\n right: `Write`: i2c::write unexpected mode"`
```

(however, i have the very odd experience that _without_ my PR it builds fine locally but with it it also fails with `cargo +nightly test --all-features`... i can't explain that right now 😕)

dbrgn:
Yeah, nightly apparently changed the assertion format. If we adapt to nightly, then tests on stable will fail.

I think I'll simply reduce the assertions to a smaller subset of the message.

dbrgn:
@rursprung should be fixed on `main`, could you rebase?

rursprung:
> @rursprung should be fixed on `main`, could you rebase?

rebased, works fine now. thanks for fixing this!

rursprung:
@dbrgn: while it wasn't strictly necessary (would've merged just fine also without this) i've now rebased this on top of the latest `main` and also adjusted a comment which still had a link to the e-h beta documentation.
once this PR lands in a release of e-h-m (it was my hope that it'd land in the release you just did) i'll be able to point to your release again in my crate (since i need this)

thanks for your work on this crate!

src/eh1/pwm.rs
```diff
@@ -0,0 +1,130 @@
...
+//! use eh1::pwm::SetDutyCycle;
+//!
+//! let err = MockError::Io(ErrorKind::NotConnected);
+//!
+//! // Configure expectations
+//! let expectations = [
+//!     PwmTransaction::get_max_duty_cycle(100),
+//!     PwmTransaction::set_duty_cycle(50),
+//!     PwmTransaction::set_duty_cycle(101).with_error(err.clone()),
+//! ];
```
dbrgn, line 18: Since the error is only used once, this would be less confusing to read:

```suggestion
//! // Configure expectations
//! let expectations = [
//!     PwmTransaction::get_max_duty_cycle(100),
//!     PwmTransaction::set_duty_cycle(50),
//!     PwmTransaction::set_duty_cycle(101).with_error(MockError::Io(ErrorKind::NotConnected)),
//! ];
```

src/eh1/pwm.rs
```diff
@@ -0,0 +1,130 @@
...
+/// MockPwm transaction kind
+#[derive(PartialEq, Clone, Debug)]
+pub enum TransactionKind {
+    /// [`SetDutyCycle::get_max_duty_cycle`] which will return the defined duty.
+    GetMaxDutyCycle(u16),
+    /// [`SetDutyCycle::set_duty_cycle`] with the expected duty.
+    SetDutyCycle(u16),
+}
+
+/// Mock SetDutyCycle implementation
```
dbrgn, line 84:
```suggestion
/// Mock PWM `SetDutyCycle` implementation
```

src/eh1/pwm.rs
```diff
@@ -0,0 +1,130 @@
...
+            .next()
+            .expect("no expectation for get_max_duty_cycle call");
+
+        assert_eq!(err, None, "error not supported by get_max_duty_cycle!");
+
+        if let TransactionKind::GetMaxDutyCycle(duty) = kind {
+            duty
+        } else {
+            unreachable!();
+        }
```
dbrgn, line 111: This branch is not actually unreachable, it could be `TransactionKind::SetDutyCycle(...)`. In that case, there _is_ a panic, but the assertion message is very user un-friendly ("unreachable code reached" instead of "expected x, found y".

I'd suggest this:

```suggestion
        match kind {
            TransactionKind::GetMaxDutyCycle(duty) => duty,
            other => panic!("expected get_max_duty_cycle, got {:?}", other),
        }
```

dbrgn (changes requested):

src/eh1/pwm.rs
```diff
@@ -0,0 +1,130 @@
...
+//! // Run and test
+//! pwm.set_duty_cycle_percent(50).unwrap();
+//! pwm.set_duty_cycle(101).expect_err("expected error return");
+//!
+//! pwm.done();
+//!
+//! // Update expectations
+//! pwm.expect(&[]);
+//! // ...
+//! pwm.done();
```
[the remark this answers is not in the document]
rursprung: i guess that was copy & paste from `pin.rs` - just noticed that the others also don't have that.

rursprung:
thanks for all your feedback @dbrgn, i've incorporated it now

dbrgn (approved):
Thanks for the updates!

dbrgn:
Released in 0.10.0-rc.2.

PR #89, opened 2023-09-14
Add support for ToggleableOutputPin for pin::Mock

### LINKED ISSUE #30: StatefulOutputPin and ToggleableOutputPin
Would you be interested in a PR that implements [ToggleableOutputPin](https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/v2/trait.ToggleableOutputPin.html)?

If I understand things correctly, the `Transaction` already knows it's state

https://github.com/dbrgn/embedded-hal-mock/blob/master/src/pin.rs#L51

so it should be relatively easy to implement [StatefulOutputPin](https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/v2/trait.StatefulOutputPin.html), which in turn would in turn make [implementing toggleable](https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/v2/toggleable/index.html) relatively trivial.

cdunster:
This PR adds support for the [`ToggleableOutputPin`](https://docs.rs/embedded-hal/latest/embedded_hal/digital/v2/trait.ToggleableOutputPin.html) trait from the `embedded-hal` crate.

Possibly closes #30, although that issue mentions wanting to do this in a stateful way but currently, the mocks don't seem to work that way. Instead, they just work on the transactions and have no internal pin state.

### DIFF
--- a/src/eh1/digital.rs
+++ b/src/eh1/digital.rs
@@ -1,15 +1,16 @@
-//! Mock digital [`InputPin`] and [`OutputPin`] implementations
+//! Mock digital [`InputPin`], [`OutputPin`], and [`StatefulOutputPin`] implementations
 //! Also mock calls to [`Wait`], assuming the `embedded-hal-async` feature is enabled.
 //!
 //! [`InputPin`]: https://docs.rs/embedded-hal/1/embedded_hal/digital/trait.InputPin.html
 //! [`OutputPin`]: https://docs.rs/embedded-hal/1/embedded_hal/digital/trait.OutputPin.html
+//! [`StatefulOutputPin`]: https://docs.rs/embedded-hal/1/embedded_hal/digital/trait.StatefulOutputPin.html
 //! [`Wait`]: https://docs.rs/embedded-hal-async/1/embedded_hal_async/digital/trait.Wait.html
 //!
 //! ```
 //! # use eh1 as embedded_hal;
 //! use std::io::ErrorKind;
 //!
-//! use embedded_hal::digital::{InputPin, OutputPin};
+//! use embedded_hal::digital::{InputPin, OutputPin, StatefulOutputPin};
 //! use embedded_hal_mock::eh1::{
 //!     digital::{Mock as PinMock, State as PinState, Transaction as PinTransaction},
 //!     MockError,
@@ -23,6 +24,9 @@
 //!     PinTransaction::get(PinState::High),
 //!     PinTransaction::set(PinState::Low),
 //!     PinTransaction::set(PinState::High).with_error(err.clone()),
+//!     PinTransaction::get_state(PinState::High),
+//!     PinTransaction::toggle(),
+//!     PinTransaction::get_state(PinState::Low),
 //! ];
 //!
 //! // Create pin
@@ -35,6 +39,10 @@
 //! pin.set_low().unwrap();
 //! pin.set_high().expect_err("expected error return");
 //!
+//! pin.is_set_high().unwrap();
+//! pin.toggle().unwrap();
+//! pin.is_set_low().unwrap();
+//!
 //! pin.done();
 //!
 //! // Update expectations
@@ -44,7 +52,7 @@
 //! ```
 
 use eh1 as embedded_hal;
-use embedded_hal::digital::{ErrorType, InputPin, OutputPin};
+use embedded_hal::digital::{ErrorType, InputPin, OutputPin, StatefulOutputPin};
 
 use crate::{common::Generic, eh1::error::MockError};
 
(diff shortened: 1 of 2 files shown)


dbrgn:
@cdunster hi, there are new conflicts introduced by other merges. Would you mind updating your branch? Thanks.

cdunster:
Hi @dbrgn, done. Thanks for looking into this PR.

cdunster:
Hi @dbrgn and happy new year, I've updated this PR to work with the new `StatefulOutputPin` that replaces `ToggleableOutputPin` in embedded-hal v1. The CI is passing so it would be nice to get this merged, let me know if I can help. Thanks.

src/eh1/pin.rs
```diff
@@ -105,6 +123,10 @@ pub enum TransactionKind {
     Set(State),
     /// Get the pin state
     Get(State),
+    /// Toggle for an expected toggle of the pin
+    Toggle,
+    /// Get the set state of the stateful pin
+    GetSetState(State),
```
dbrgn, line 129: Should we rename this to `GetState`? Otherwise users could be confused: Will it get or set the state? 😄 

The factory method would also need a rename to `get_set_state`.

src/eh1/pin.rs
```diff
@@ -199,16 +221,73 @@ impl InputPin for Mock {
...
+            "expected pin::is_set_high"
+        );
+
+        if let Some(e) = err {
+            Err(e)
+        } else if let TransactionKind::GetSetState(v) = kind {
+            Ok(v == State::High)
+        } else {
+            unreachable!();
+        }
```
dbrgn, line 256: Hmm, I'm not sure if this makes sense. In the `assert_eq!` above, you're asserting that the transaction is `GetSetState(High)`. This already implies that the state is high, so the check at 252 doesn't do anything meaningful.

On the other hand, shouldn't this transaction work both for high and low states? In the case of a low state, it would simply return `Ok(false)`.

I suspect that the `assert_eq!` needs to be replaced with something like `assert!(matches!(kind, TransactionKind::GetSetState(_)))` and an appropriate error message. Same thing for `is_set_low`. In that case, no matter whether the driver calls `is_set_high` or `is_set_low`, it will work (and return either true or false).

src/eh1/pin.rs
```diff
@@ -249,4 +328,30 @@ mod test {
...
+    #[test]
+    fn test_stateful_output_pin() {
+        let expectations = [
+            Transaction::new(GetSetState(State::Low)),
+            Transaction::new(Toggle),
+            Transaction::get_set_state(State::High),
+            Transaction::toggle(),
+            Transaction::new(GetSetState(State::Low))
+                .with_error(MockError::Io(ErrorKind::NotConnected)),
+            Transaction::new(Toggle).with_error(MockError::Io(ErrorKind::NotConnected)),
```
dbrgn, line 341: Why not use the factory methods instead of passing in the transaction kinds directly?

dbrgn (changes requested):

src/eh1/pin.rs
cdunster: I kinda feel like there is no good name for this :sweat_smile:. I thought that `GetState(State)` might be confused with `Get(State)` so I opted for this. I'm happy to change it though, I could also do `GetTheSetState` or `GetCurrentSetState` but they're both a bit verbose, maybe `GetCurrentState`? :shrug:

src/eh1/pin.rs
cdunster: I was just testing both ways for completion and this is the way that the other tests define transactions so I followed suit.

src/eh1/pin.rs
cdunster: Oh yeah, good catch! Thanks. I've added a `fixup` commit to fix it and added some unit tests for it too.

src/eh1/pin.rs
dbrgn: I think `GetState` is fine. If you prefer `GetCurrentState`, that would be OK for me too.

cdunster: I went for `GetState` in the end.

src/eh1/pin.rs
dbrgn: Thanks! Fixup commits are fine as well, I can squash them before merging 🙂

dbrgn:
Thanks for the updates! I'll rebase this branch against `main`. The `pin` module for eh1 was renamed to `digital` in #113, so I'll have to adjust the second commit for that.

src/eh1/digital.rs
```diff
@@ -132,6 +150,10 @@ pub enum TransactionKind {
     Set(State),
     /// Get the pin state
     Get(State),
+    /// Toggle for an expected toggle of the pin
```
dbrgn, line 153: That sentence looks a bit strange... Maybe a relict from some rewording?

I'll change to `Toggle the pin state`.

dbrgn (approved):
LGTM, thanks! Rebase was successful. I'll fix that remaining wording issue myself, and will then merge.

PR #91, opened 2023-09-18
Add support for embedded_hal_async::digital::Wait

cdunster:
This PR adds support for the async [`Wait`](https://docs.rs/embedded-hal-async/0.2.0-alpha.1/embedded_hal_async/digital/trait.Wait.html) trait from the `embedded-hal-async` crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ me know.
 | Actual delay                                | ✅           | ✅               |
 | Serial                                      | ✅           | -                |
 | RNG                                         | -            | -                |
-| I/O pins (including PWM)                    | ✅           | -                |
+| I/O pins (including PWM)                    | ✅           | ✅               |
 | ADC                                         | ✅           | -                |
 | Timers (with `embedded-time` Cargo feature) | ✅           | -                |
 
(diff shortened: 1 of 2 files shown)


dbrgn:
@cdunster hi, there are new conflicts introduced by other merges. Would you mind updating your branch? Thanks.

cdunster:
Hi @dbrgn, I've updated my branch. Thanks for looking into this PR.

avsaase:
I gave this PR a try and it works well. Would be possible to add a configurable delay before these async functions resolve? In my code I use timeouts for control flow so I need a way to make these futures not resolve immediately.

dbrgn:
@cdunster looking great, thanks!

Regarding the configurable delay: If such a delay would be added, is that something that should be added to other async mock method calls as well, not just digital pins?

On the other hand, this method is explicitly about waiting, so a configurable wait time might make sense. However, the point about the tokio dependency is a valid one. Ideally, we'd have an executor-independent async sleep implementation, but I think that's not currently possible in Rust, right?

In any case, I'll rebase and merge this PR for now. I'll tag a new release soon, but more features (even breaking changes) can always be added. If you're interested in a configurable delay, feel free to open a new PR, where it can be discussed.

PR #92, opened 2023-10-17
eh0: add pwm mock

schphil:
This is an initial draft for adding a PWM mock in eh0. I'm not sure if differentiation between multiple channels needs to be supported directly.  Any feedback would be greatly appreciated.

### DIFF
--- a/src/eh0.rs
+++ b/src/eh0.rs
@@ -12,6 +12,7 @@ pub mod delay;
 pub mod i2c;
 pub mod pin;
 pub mod serial;
+pub mod pwm;
 pub mod spi;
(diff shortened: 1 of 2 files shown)


dbrgn:
Hi, thanks for the PR!

I wonder how much sense it makes to add new PWM mocks for eh0 if eh1 is nearing completion... A mock for eh1 `SetDutyCycle` was just merged in #88. What do you think, is this something you'd need/use for quite a while, or would it be replaced very soon?

dbrgn:
Closing for now, feel free to leave a comment if you think it's still important.

Issue #100, opened 2023-12-23
IO implementation for embedded-io


Artur-Romaniuk:
With the release of `embedded-hal 1.0.0`  `Serial` trait has been removed in favor of using `embedded-io`. This issue proposes adding IO mocks as a preferred method over existing 'Serial' which is based on  'embedded-hal-nb'.

PR #101, opened 2023-12-23
IO implementation for embedded-io

Artur-Romaniuk:
Related to #100


newAM (changes requested):
Looks good, thanks for adding the feature, just some structural changes I would make.

Cargo.toml
```diff
@@ -22,20 +22,24 @@ rust-version = "1.63"
...
-embedded-hal-async = ["dep:embedded-hal-async"]
+embedded-hal-async = ["dep:embedded-hal-async", "dep:embedded-io-async"]
 
 default = ["eh0", "embedded-time"]
 
 [dependencies]
-eh0 = { package = "embedded-hal", version = "0.2.7", features = ["unproven"], optional = true }
+eh0 = { package = "embedded-hal", version = "0.2.7", features = [
+    "unproven",
+], optional = true }
```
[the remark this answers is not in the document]
Artur-Romaniuk: This is a result of automatic formating on file save. If you could tell me if this file gets formated the same way for you I would be thankful. If it happens only on my machine I will revert this change.

Cargo.toml
```diff
@@ -22,20 +22,24 @@ rust-version = "1.63"
 
 [features]
 eh0 = ["dep:eh0", "dep:nb"]
-eh1 = ["dep:eh1", "dep:embedded-hal-nb"]
+eh1 = ["dep:eh1", "dep:embedded-hal-nb", "dep:embedded-io"]
```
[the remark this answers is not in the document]
Artur-Romaniuk: Ok, but I guess README should be also updated to better describe project structure.
I think that it is reasonable to say that embedded-io is just as much emedded-hal as embedded-hal-nb is. All of these are parts of the core HALs and some of them like -io and -can got separated, but in that regard (outdated and maybe soon deprecated) -nb hal is also a separate repo.
1.0 argument is indeed compelling.

Other than that I don't have strong opinions on this and I can move it behind a separate feature gate if decided so.

Artur-Romaniuk (commented):
Will apply suggested changes.

Cargo.toml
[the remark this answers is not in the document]
dbrgn: What editor and formatting plugin do you use? rustfmt does not reformat TOML files by default.

I would prefer one-line-per-dependency (I find it easier to read), so it would be nice if you could revert the change 🙂

Cargo.toml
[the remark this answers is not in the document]
dbrgn: I tend to support @Artur-Romaniuk's argument that embedded-io could be considered part of embedded-hal.

The 1.0 argument is a good one, but OTOH embedded-hal-mock isn't at 1.0 yet, and thus we can freely include version bumps of the -io dependencies in minor updates.

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+    Seek(SeekFrom),
+    /// WriteReady transaction
+    WriteReady(bool),
+    /// ReadReady transaction
+    ReadReady(bool),
+    /// FillBuff transaction
+    FillBuff,
+    /// Consume transaction
+    Consume(usize),
+}
```
dbrgn, line 74: As a thought: Might it be better for readability if the variants with associated data (e.g. `WriteReady(bool)`) would use named parameters (e.g. `WriteReady { ready: bool }`)?

I have no strong opinon on this, so if you prefer the brevity, that's fine too. (After all, this type is usually instantiated through the constructor methods of the transaction anyways.)

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+
+impl Read for Mock {
+    fn read(&mut self, buffer: &mut [u8]) -> Result<usize, Self::Error> {
+        let w = self.next().expect("no expectation for io::read call");
+        assert_eq!(w.expected_mode, Mode::Read, "io::read unexpected mode");
+        buffer.copy_from_slice(&w.response);
+
+        match w.expected_err {
+            Some(err) => Err(err),
+            None => Ok(buffer.len()),
```
dbrgn, line 225: Is this correct in the case where `w.response` is shorter than `buffer`? Shouldn't it be the shorter of the two lengths?

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+        );
+
+        match w.expected_err {
+            Some(err) => Err(err),
+            None => Ok(buffer.len()),
+        }
+    }
+
+    fn flush(&mut self) -> Result<(), Self::Error> {
+        let w = self.next().expect("no expectation for io::flush call");
```
dbrgn, line 207: The letter `w` was probably chosen in the `write` method to stand for "write". But in all other impls it seems a bit weird.

Maybe use `t` or `transaction` instead? (Here, and in all other impls as well.)

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+        if let Mode::Seek(expected_pos) = w.expected_mode {
+            assert_eq!(expected_pos, pos, "io::seek unexpected mode");
+
+            let ret_offset: u64 = u64::from_be_bytes(w.response.try_into().unwrap());
+            match w.expected_err {
+                Some(err) => Err(err),
+                None => Ok(ret_offset),
+            }
+        } else {
+            panic!("unexpected seek mode");
```
dbrgn, line 243: As a user of the library, this can lead to hard-to-debug test failures.

This panic happens if the test expects a seek transaction, but some other action is triggered by the code.

Something like this might be much better:

```rust
panic!("expected seek transaction, but instead encountered {:?}", w.expected_mode);
```

(Same thing for all other impls.)

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+        let response_vec = w.response;
+
+        match w.expected_err {
+            Some(err) => Err(err),
+
+            // SAFETY : This is a memory leak. This is done on purpose to allow for a return
+            // of a slice (pointer) to the internal buffer, which doesn't exist in mock implementation.
+            // This way, after each call, response is going to be put on the heap and ref is going to be returned, without freeing this memory.
+            // For mocking purposes this should be an acceptable solution.
+            None => Ok(Box::leak(response_vec.into_boxed_slice())),
```
dbrgn, line 292: I think this solution is fine for a mocking library.

However, another option would be to add a buffer to the mock. Since most mocks use an alias to `Generic<T>`, this isn't trivial to add. But one way might be to parametrize the `Generic` type and add a general-purpose data field, something like this:

```rust
#[derive(Debug, Clone)]
pub struct Generic<T: Clone + Debug + PartialEq, B = ()> {
    expected: Arc<Mutex<VecDeque<T>>>,
    done_called: Arc<Mutex<DoneCallDetector>>,
    data: B,
}
```

This way, a mock could add associated data to the generic mock. In this case, it would be a generic buffer of a certain size. But it could also be a struct, depending on the mock type.

src/eh1/io.rs
```diff
@@ -0,0 +1,617 @@
...
+        }
+    }
+
+    fn consume(&mut self, amt: usize) {
+        let w = self.next().expect("no expectation for io::consume call");
+
+        match w.expected_mode {
+            Mode::Consume(expected_amt) if w.expected_err.is_none() => {
+                assert_eq!(expected_amt, amt, "io::consume unexpected amount");
+            }
```
dbrgn, line 302: What about the case where `expected_err` is some?

Cargo.toml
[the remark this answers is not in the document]
Artur-Romaniuk: Yup, this is a result of formatting from "Even Better TOML" extension in VS Code. Reverting this change.

src/eh1/io.rs
Artur-Romaniuk: Added explicit panic for an expected_err as it is a user error to expect error for an infallible command `consume`.

Artur-Romaniuk:
Rebased on main. Would love to have some feedback.

src/common.rs
```diff
@@ -99,10 +101,20 @@ where
...
         assert!(e.is_empty(), "Not all expectations consumed");
     }
+
+    /// Get the mock data
+    pub fn mock_data(&self) -> &Option<Data> {
+        &self.mock_data
+    }
+
+    /// Set the mock data
+    pub fn set_mock_data(&mut self, data: Option<Data>) {
```
newAM, line 111: Would be nice to have some example uses here.

newAM (approved):
I had some nitpicks with the documentation links but I just did the changes and pushed them since that's easier.

Looks good to me overall.  We're going to have unfortunate conflicts with the other PR that touches delays, but that's unavoidable.  Will wait for dbrgn's feedback on this too.

src/common.rs
Artur-Romaniuk: Actually, I wasn't sure myself on how to do this properly. Getters and setter feel kinda wrong here, but sharing mut ref can be problematic as well.

Not sure on how to show examples for a getter and setter methods, though.

Added an appropriate comment on top of the module, that describes mock_data functionality.

Artur-Romaniuk:
This PR went stale for a while now. @dbrgn could take a look?

dbrgn:
@Artur-Romaniuk because this PR was created on top of the delay branch (#104), I wanted to wait until that branch is ready and merged (and left feedback there) to make review easier.

If you want to make PRs easier to review, try to keep them small and focussed. Having a lot of practice with git rebase (especially interactive rebase) helps a lot with that.

Artur-Romaniuk:
> @Artur-Romaniuk because this PR was created on top of the delay branch (#104), I wanted to wait until that branch is ready and merged (and left feedback there) to make review easier.
> 
> If you want to make PRs easier to review, try to keep them small and focussed. Having a lot of practice with git rebase (especially interactive rebase) helps a lot with that.

That was a fuckup...
Must have rebased this branch on wrong origin/main. Force pushed rebased version with all unecessary commits dropped.

src/eh1/io.rs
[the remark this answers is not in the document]
dbrgn: Yeah, I think the suggestion makes sense.

src/eh1/io.rs
dbrgn: Fine for me.

Cargo.toml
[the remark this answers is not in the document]
dbrgn: Note: In case `embedded-io` is causing troubles with its versioning, we can always extract it into its own feature.

src/common.rs
```diff
@@ -20,26 +20,32 @@ use std::{
 /// Note that the implementation uses an `Arc<Mutex<...>>` internally, so a
 /// cloned instance of the mock can be used to check the expectations of the
 /// original instance that has been moved into a driver.
+///
+/// Optional Data type allows for storage of arbitrary state data in mocks.
+/// Mock data can be accessed via appropriate getter/setter methods, i.e. `mock_data()` and `set_mock_data()`.
```
dbrgn, line 25:
```suggestion
/// The `mock_data` field allows for storage of arbitrary state data in mocks.
/// Mock data can be accessed via appropriate getter/setter methods, i.e. `mock_data()` and `set_mock_data()`.
```

src/common.rs
```diff
@@ -20,26 +20,32 @@ use std::{
...
+///
 #[derive(Debug, Clone)]
-pub struct Generic<T: Clone + Debug + PartialEq> {
+pub struct Generic<T: Clone + Debug + PartialEq, Data = ()> {
```
dbrgn, line 28: Can you rename `Data` to `TData`?

src/common.rs
dbrgn: I think those docs should suffice 👍🏻

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
+//! IO mock implementations.
+//!
+//! This mock supports the specification and checking of expectations to allow
+//! automated testing of IO based drivers. Mismatches between expected and
```
dbrgn, line 4:
```suggestion
//! automated testing of `embedded-io` based drivers. Mismatches between expected and
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+//!     IoTransaction::write(vec![10, 20]),
+//!     IoTransaction::read(vec![20, 30]),
+//!     IoTransaction::flush(),
+//! ];
+//!
+//! let mut io = IoMock::new(&expectations);
+//!
+//! // Writing
+//! let ret = io.write(&[10, 20]).unwrap();
+//! assert_eq!(ret, 2);
```
dbrgn, line 25:
```suggestion
//! let bytes_written = io.write(&[10, 20]).unwrap();
//! assert_eq!(bytes_written, 2);
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+//!
+//! // Reading
+//! let mut buffer = [0; 2];
+//! let ret = io.read(&mut buffer).unwrap();
+//! assert_eq!(buffer, [20, 30]);
+//! assert_eq!(ret, 2);
```
dbrgn, line 31:
```suggestion
//! let bytes_read = io.read(&mut buffer).unwrap();
//! assert_eq!(buffer, [20, 30]);
//! assert_eq!(bytes_read, 2);
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+//! // Finalizing expectations
+//! io.done();
+//!
+//! // optional async
+//! #[cfg(feature = "embedded-hal-async")]
+//! async {
+//!     use embedded_io_async;
+//!     let mut io = IoMock::new(&[IoTransaction::write(vec![10, 20])]);
+//!     let ret = embedded_io_async::Write::write(&mut io, &[10, 20]).await.unwrap();
+//!     assert_eq!(ret, 2);
```
dbrgn, line 45:
```suggestion
//!     let bytes_written = embedded_io_async::Write::write(&mut io, &[10, 20]).await.unwrap();
//!     assert_eq!(bytes_written, 2);
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+//! assert_eq!(buffer, [20, 30]);
+//! assert_eq!(ret, 2);
+//!
+//! // Flushing
+//! io.flush().unwrap();
+//!
+//! // Finalizing expectations
+//! io.done();
+//!
+//! // optional async
```
dbrgn, line 39:
```suggestion
//! // Async is supported as well, if the `embedded-hal-async` feature is enabled
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// transaction mode is correct prior to returning the error.
+    expected_err: Option<ErrorKind>,
+}
+
+impl Transaction {
+    /// Create a write transaction
+    pub fn write(expected: Vec<u8>) -> Transaction {
+        Transaction {
+            expected_mode: Mode::Write,
+            expected_data: expected,
```
dbrgn, line 98:
```suggestion
    pub fn write(expected_data: Vec<u8>) -> Transaction {
        Transaction {
            expected_mode: Mode::Write,
            expected_data,
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+            transaction.expected_mode,
+            Mode::Read,
+            "io::read unexpected mode"
+        );
+
+        if transaction.response.len() > buffer.len() {
+            panic!("response longer than read buffer for io::read");
+        }
+
+        let len = std::cmp::min(buffer.len(), transaction.response.len());
```
dbrgn, line 240: Since we already assert above, that the response length can't be larger than the buffer length, can't we just use the response length here?

```suggestion
        let len = transaction.response.len();
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+        }
+    }
+}
+
+impl Seek for Mock {
+    fn seek(&mut self, pos: SeekFrom) -> Result<u64, Self::Error> {
+        let transaction = self.next().expect("no expectation for io::seek call");
+
+        if let Mode::Seek(expected_pos) = transaction.expected_mode {
+            assert_eq!(expected_pos, pos, "io::seek unexpected mode");
```
dbrgn, line 255:
```suggestion
            assert_eq!(expected_pos, pos, "io::seek unexpected pos");
```

src/eh1/io.rs
```diff
@@ -0,0 +1,684 @@
...
+    fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
+        let transaction = self.next().expect("no expectation for io::fill_buf call");
+        assert_eq!(
+            transaction.expected_mode,
+            Mode::FillBuff,
+            "io::fill_buf unexpected mode"
+        );
+
+        let response_vec = transaction.response;
+        self.set_mock_data(Some(response_vec));
```
dbrgn, line 317: Since this variable is only used once, I think it can be inlined.

```suggestion
        self.set_mock_data(Some(transaction.response;));
```

dbrgn (changes requested):
I left a few last comments. The rest is looking great, I think once the comments are addressed this can be merged.

victorbnl:
Any chance this gets fixed and merged in a near future?

dbrgn:
@Artur-Romaniuk do you have plans to finish this PR, or would you like for someone else to take over?

esden:
Would like to bump this issue. Any chance to get embedded-io support added, maybe even async?

PR #104, opened 2023-12-27
Added async delay and checked delay

Artur-Romaniuk:
Delay lacked async implementation.

Needed it for my test so I added it.

### DIFF
--- a/src/eh1/delay.rs
+++ b/src/eh1/delay.rs
@@ -1,5 +1,5 @@
 //! Delay mock implementations, implementing both sync and async
-//! [`BlockingDelay`](https://docs.rs/embedded-hal/latest/embedded_hal/delay/trait.BlockingDelay.html)
+//! [`DelayNs`](https://docs.rs/embedded-hal/latest/embedded_hal/delay/trait.DelayNs.html)
 //! traits.
 //!
 //! ## Choosing a Delay Implementation
@@ -74,9 +74,9 @@ pub struct Transaction {
 }
 
 /// Nanoseconds per microsecond
-const NANOS_PER_MICRO: u64 = 1_000;
+const NANOS_PER_US: u64 = 1_000;
 /// Nanoseconds per millisecond
-const NANOS_PER_MILLI: u64 = 1_000_000;
+const NANOS_PER_MS: u64 = 1_000_000;
 
 impl Transaction {
     /// Create a new delay transaction
@@ -87,52 +87,52 @@ impl Transaction {
         }
     }
 
-    /// Crete a new delay_ns transaction
+    /// Create a new delay_ns transaction
     pub fn delay_ns(ns: u32) -> Transaction {
-        Transaction::new(TransactionKind::Delay(ns.into()))
+        Transaction::new(TransactionKind::DelayNs(ns.into()))
     }
 
-    /// Crete a new delay_us transaction
+    /// Create a new delay_us transaction
     pub fn delay_us(us: u32) -> Transaction {
-        Transaction::new(TransactionKind::Delay(us as u64 * NANOS_PER_MICRO))
+        Transaction::new(TransactionKind::DelayNs(us as u64 * NANOS_PER_US))
     }
 
-    /// Crete a new delay_ms transaction
+    /// Create a new delay_ms transaction
     pub fn delay_ms(ms: u32) -> Transaction {
-        Transaction::new(TransactionKind::Delay(ms as u64 * NANOS_PER_MILLI))
+        Transaction::new(TransactionKind::DelayNs(ms as u64 * NANOS_PER_MS))
     }
 
     /// Create a new blocking delay_ns transaction
     pub fn blocking_delay_ns(ns: u32) -> Transaction {
-        Transaction::new(TransactionKind::BlockingDelay(ns.into()))
+        Transaction::new(TransactionKind::BlockingDelayNs(ns.into()))
     }
 
-    /// Crete a new blocking delay_us transaction
+    /// Create a new blocking delay_us transaction
     pub fn blocking_delay_us(us: u32) -> Transaction {
-        Transaction::new(TransactionKind::BlockingDelay(us as u64 * NANOS_PER_MICRO))
+        Transaction::new(TransactionKind::BlockingDelayNs(us as u64 * NANOS_PER_US))
     }
 
     /// Create new blocking delay_ms transaction
     pub fn blocking_delay_ms(ms: u32) -> Transaction {
-        Transaction::new(TransactionKind::BlockingDelay(ms as u64 * NANOS_PER_MILLI))
+        Transaction::new(TransactionKind::BlockingDelayNs(ms as u64 * NANOS_PER_MS))
     }
 
-    /// Crete a new async delay_ns transaction
+    /// Create a new async delay_ns transaction
     #[cfg(feature = "embedded-hal-async")]
     pub fn async_delay_ns(ns: u32) -> Transaction {
-        Transaction::new(TransactionKind::AsyncDelay(ns.into()))
+        Transaction::new(TransactionKind::AsyncDelayNs(ns.into()))
     }
 
-    /// Crete a new async delay_us transaction
+    /// Create a new async delay_us transaction
     #[cfg(feature = "embedded-hal-async")]
     pub fn async_delay_us(us: u32) -> Transaction {
-        Transaction::new(TransactionKind::AsyncDelay(us as u64 * NANOS_PER_MICRO))
+        Transaction::new(TransactionKind::AsyncDelayNs(us as u64 * NANOS_PER_US))
     }
 
-    /// Crete a new async delay_ms transaction
+    /// Create a new async delay_ms transaction
     #[cfg(feature = "embedded-hal-async")]
     pub fn async_delay_ms(ms: u32) -> Transaction {
-        Transaction::new(TransactionKind::AsyncDelay(ms as u64 * NANOS_PER_MILLI))
+        Transaction::new(TransactionKind::AsyncDelayNs(ms as u64 * NANOS_PER_MS))
     }
 
     /// Perform an actual delay for this transaction
@@ -145,12 +145,14 @@ impl Transaction {
 /// MockDelay transaction kind.
 #[derive(PartialEq, Eq, Clone, Debug)]
 pub enum TransactionKind {
-    /// Any delay in nanoseconds, blocking or async. Should be a default kind for most use cases.
-    Delay(u64),
+    /// Expect any type of delay in nanoseconds.
+    ///
+    /// The delay may be either blocking or async. In most cases, this is what you'll want to use.
+    DelayNs(u64),
     /// Delay in nanoseconds, must be a blocking delay. Expectation will fail for async delays.
-    BlockingDelay(u64),
+    BlockingDelayNs(u64),
     /// Delay in nanoseconds, must be an async delay. Expectation will fail for blocking delays.
-    AsyncDelay(u64),
+    AsyncDelayNs(u64),
 }
 
 /// Mock Delay implementation with checked calls
@@ -168,10 +170,10 @@ impl delay::DelayNs for CheckedDelay {
         let transaction = self.next().expect("no expectation for delay call");
 
         match transaction.kind {
-            TransactionKind::BlockingDelay(n) => assert_eq!(n, ns.into(), "wrong delay value"),
-            TransactionKind::Delay(n) => assert_eq!(n, ns.into(), "wrong delay value"),
+            TransactionKind::BlockingDelayNs(n) => assert_eq!(n, ns.into(), "wrong delay value"),
+            TransactionKind::DelayNs(n) => assert_eq!(n, ns.into(), "wrong delay value"),
             _ => panic!(
-                "Wrong kind of delay. Expected Delay or BlockingDelay got {:?}",
+                "Wrong kind of delay. Expected DelayNs or BlockingDelayNs got {:?}",
                 transaction.kind
             ),
         }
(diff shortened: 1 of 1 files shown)


src/eh1/delay.rs
```diff
@@ -68,12 +67,50 @@ impl delay::DelayNs for StdSleep {
     fn delay_ns(&mut self, ns: u32) {
         thread::sleep(Duration::from_nanos(ns as u64));
     }
+}
+
+#[cfg(feature = "embedded-hal-async")]
+impl embedded_hal_async::delay::DelayNs for StdSleep {
+    async fn delay_ns(&mut self, _ns: u32) {
+        // no-op
```
dbrgn, line 75: This one shouldn't be a no-op, right?

src/eh1/delay.rs
```diff
@@ -68,12 +67,50 @@ impl delay::DelayNs for StdSleep {
...
+    }
+}
+
+#[cfg(test)]
+mod test {
```
dbrgn, line 80: The tests below don't actually test anything, except that the functions don't panic. That's also, why the issue above wasn't caught.

Suggestion:

- In the noop tests, call `delay_ms(500)` and assert that less than 100 ms passed (by measuring time before and after)
- In std tests, call `delay_ms(100)` and assert that more than 100 ms passed (by measuring time before and after)

Use https://doc.rust-lang.org/std/time/struct.Instant.html as monotonic time source.

dbrgn (changes requested):
Thanks! I left two comments.

Artur-Romaniuk:
@dbrgn Fixed raised issues.

At the same time, the existing implementations weren't actually mocks, rather stubs. I added a new delay instance called CheckedDelay which is a proper mock with transactions.

dbrgn:
Hi @Artur-Romaniuk, thanks for the update! I improved the module documentation and pushed a commit directly to your fork, I hope that's OK. I'll also rebase the branch against `main`, because there we changed the testing configuration (causing the tests to fail right now).

dbrgn:
I'm really confused why the doc tests fail in CI... They work without any issues locally 😕 I'll try to investigate another day (but any pointers would be welcome).

By the way, @Artur-Romaniuk, I noticed that you used a different e-mail for the first commit compared to the others. If this was a mistake, now you could still fix it using git rebase 🙂

Artur-Romaniuk:
Happy to see your changes. Will look at it tomorrow.

src/eh1/delay.rs
```diff
@@ -1,23 +1,403 @@
...
+//!     Transaction::delay_ns(50_000_000),
+//!     Transaction::delay_us(60_000),
+//!     Transaction::delay_ms(70).wait(),
+//! ];
+//!
+//! let mut delay = CheckedDelay::new(&transactions);
+//!
+//! delay.delay_ms(50); // Conversion to nanoseconds
+//! delay.delay_ms(60); // Conversion to microseconds
+//! delay.delay_ms(70); // This will actually delay
```
dbrgn, line 53: I had to read the example a few times to understand it. If I understand it correctly, the transactions offer different constructor methods, but in the end the trait only tests for `delay_ns` calls, so they are all interchangeable.

The `delay.delay_ms(...)` calls would be done by the driver-under-test.

This makes sense to test that the implementation of the mock works. However, with the comments you added ("Conversion to nanoseconds / microseconds") it looks as if we would assert that the driver actually calls the `delay_us` and `delay_ns` methods... However, that's not the case.

Maybe my confusion also comes from the fact that the transaction types differentiate by ms/us/ns.

Can we have simpler transactions, and remove `DelayMs` and `DelayUs` in favor of `DelayNs`? We can do the nanosecond calculations directly in the constructor methods (i.e. `pub fn delay_ms(...)` would construct a `DelayNs` transaction).

src/eh1/delay.rs
```diff
@@ -1,23 +1,403 @@
...
+    AsyncDelayUs(u32),
+    /// Any delay in microseconds, blocking or async
+    AnyDelayUs(u32),
+    /// Delay in milliseconds
+    DelayMs(u32),
+    /// Asynchronous delay in milliseconds
+    AsyncDelayMs(u32),
+    /// Any delay in milliseconds, blocking or async
+    AnyDelayMs(u32),
+}
```
dbrgn, line 171: Why do we have Delay vs AsyncDelay vs AnyDelay?

If I understand it correctly, the Delay* variants can only deal with sync calls, the AsyncDelay* variants can only deal with async calls, and AnyDelay* forwards calls to both traits.

However, in which situation would one want a mock object that only supports one or the other variant? Couldn't we just implement both interfaces (as long as the necessary Cargo features are enabled) and call it a day?

Artur-Romaniuk: Interface implementation is something else, as this is done for a CheckedDelay struct and not a transaction.

The idea behind differentiation between blocking and async behavior is to allow tests to validate that async functions properly use delays.
E.g.
In a big async function, there should be 100ms delay. If the intention of the author is to use non-blocking delay, they can specify so in their tests. Now, if they include some external dependency or call some functions via polymorphic trait, tests would fail for a 100ms blocking delay. This allows for more specific tests, where some calls are supposed to be short and blocking and some should be long and async.

I wouldn't remove this functionality, but if this is an uncommon scenario and user-friendliness is of a big concern, then maybe these transactions can be renamed from:
-  Delay*
-  AsyncDelay*
-  AnyDelay*

to:
-  Delay* (blocking or async, don't care. sane default for 99% cases)
-  AsyncDelay* (must be async)
-  BlockingDelay* (must be blocking)

src/eh1/delay.rs
threema-danilo: Ah, I see! Yes, I think your suggestion with the big `DelayNs` value makes sense.

dbrgn: Ah, I see! Yes, I think your suggestion with the big `DelayNs` value makes sense.

src/eh1/delay.rs
dbrgn: I see! Yeah, I think this is quite a specialized use case. However, with your rename suggestion, I think that would already be a big improvement.

What's your opinion, @newAM?

newAM: I think the rename is a good idea, with maybe a note in the docstrings of each that `Delay` should be used in almost all cases.

Artur-Romaniuk: @dbrgn @newAM I've updated docstrings to state, that Delay is a preferred default for most use cases.
Also, the examples shown on top of the module use Delay.

src/eh1/delay.rs
```diff
@@ -1,5 +1,5 @@
 //! Delay mock implementations, implementing both sync and async
-//! [`DelayNs`](https://docs.rs/embedded-hal/latest/embedded_hal/delay/trait.DelayNs.html)
+//! [`BlockingDelay`](https://docs.rs/embedded-hal/latest/embedded_hal/delay/trait.BlockingDelay.html)
```
dbrgn, line 2: Was this change done by mistake? `BlockingDelay` is not a trait that exists in e-h.

I suspect this was done accidentally as part of an automated rename.

src/eh1/delay.rs
```diff
@@ -150,24 +145,12 @@ impl Transaction {
...
-    DelayMs(u32),
-    /// Asynchronous delay in milliseconds
-    AsyncDelayMs(u32),
-    /// Any delay in milliseconds, blocking or async
-    AnyDelayMs(u32),
+    Delay(u64),
+    /// Delay in nanoseconds, must be blocking
+    BlockingDelay(u64),
+    /// Delay in nanoseconds, must be async
+    AsyncDelay(u64),
```
dbrgn: Nit: While I really like the simplification, I think that for public APIs the time unit shoulud be obvious, and not just be part of the docs.

Two ideas:

- `DelayNs` / `BlockingDelayNs` / `AsyncDelayNs` (my preferred variant)
- `Delay { ns: u64 }` / `BlockingDelayNs { ns: u64 }` / `AsyncDelayNs { ns: u64 }`

src/eh1/delay.rs
```diff
@@ -89,55 +87,52 @@ impl Transaction {
         }
     }
 
-    /// Create a new delay_ns transaction
+    /// Crete a new delay_ns transaction
```
dbrgn, line 90: Typo: `Crete` -> `Create`. There are other occurrences of "crete" in this file as well.

src/eh1/delay.rs
```diff
@@ -74,11 +74,9 @@ pub struct Transaction {
 }
 
 /// Nanoseconds per microsecond
-const NANOS_PER_MICRO: u32 = 1_000;
+const NANOS_PER_MICRO: u64 = 1_000;
 /// Nanoseconds per millisecond
-const NANOS_PER_MILLI: u32 = 1_000_000;
-/// Microseconds per millisecond
-const MICROS_PER_MILLI: u32 = 1_000;
+const NANOS_PER_MILLI: u64 = 1_000_000;
```
dbrgn: Nit: I'd personally prefer `*_PER_US` and `*_PER_MS` to be aligned with the `*_delay_us` and `*_delay_ms` methods. But if you prefer the variant above, feel free to keep it.

src/eh1/delay.rs
```diff
@@ -1,23 +1,286 @@
...
+
+/// MockDelay transaction kind.
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub enum TransactionKind {
+    /// Any delay in nanoseconds, blocking or async. Should be a default kind for most use cases.
+    Delay(u64),
+    /// Delay in nanoseconds, must be a blocking delay. Expectation will fail for async delays.
+    BlockingDelay(u64),
+    /// Delay in nanoseconds, must be an async delay. Expectation will fail for blocking delays.
+    AsyncDelay(u64),
```
dbrgn, line 153: Nit: The way rustdoc works, is that on overview pages (where multiple types are listed), the first line of the docstring is shown. Once you open the detail page, you see the full description.

This means that the following formatting might make it easier to navigate (plus some minor rewording):

```suggestion
    /// Expect any type of delay in nanoseconds
    ///
    /// The delay may be either blocking or async. In most cases, this is what you'll want to use.
    Delay(u64),
    /// Expect a blocking delay in nanoseconds
    ///
    /// The delay must be blocking. Expectation will fail for async delays.
    BlockingDelay(u64),
    /// Expect an async delay in nanoseconds
    ///
    /// The delay must be async. Expectation will fail for blocking delays.
    AsyncDelay(u64),
```

dbrgn (changes requested):
Thanks for the updates, @Artur-Romaniuk! I think the simplification with the nanosecond-only delay has improved the readability a lot!

I left some more comments. I hope that was the last round of review.

(Note to self: When merging, we should use squash mode.)

dbrgn:
> Not really that relevant to this PR but something minor and relevant _enough_ that I figure I should bring it up here:
> In looking at `delay.rs` on `main` I saw that currently methods with default implementations are being overridden. AFAICT only `delay_ns` needs to be implemented. I'm wondering if there's a good reason to override `delay_ms`/`delay_us` @dbrgn

I think you are right! @Artur-Romaniuk could you test if everything keeps working if you remove those potentially unneeded overrides?

Artur-Romaniuk:
> > Not really that relevant to this PR but something minor and relevant _enough_ that I figure I should bring it up here:
> > In looking at `delay.rs` on `main` I saw that currently methods with default implementations are being overridden. AFAICT only `delay_ns` needs to be implemented. I'm wondering if there's a good reason to override `delay_ms`/`delay_us` @dbrgn
> 
> I think you are right, thanks for bringing this up! @Artur-Romaniuk could you test if everything keeps working if you remove those potentially unneeded overrides?

I think there is a valid reason for doing it this way.
The way the std version is implemented with default implementations is that they call the `delay_ns` function for N amount of times in order to achieve us or ms. This means, that the mock would only receive `delay_ns` calls.

In that case, mock would have to add N expectations of `delay_ns` for each call of `delay_ms`. Not only does it complicate the expectation addition logic, but also expectation verification becomes much harder.
Look at this example:

```
delay_ms(5);
// do something
delay_ns(10);
```

using defaults, this would equate to

```
for(some N amount of times){
   delay_ns(MAX_NS);
} 
delay_ns(5 * NANOS_PER_MILI - N*MAX_NS);
// do something
delay_ns(10);
```

I believe the current impl is the easiest one and there is no benefit in using default impl.

src/eh1/delay.rs
Artur-Romaniuk: Not really sure that I understand your comment about rustdocs. This is how both your and mine versions look to me:
image

image

image

image

Both of them look identical to me. I would suggest that you do your preferred way of formatting and just push directly to my fork.

src/eh1/delay.rs
Artur-Romaniuk: Yup, accidental rename. Fixed.

src/eh1/delay.rs
Artur-Romaniuk: While `TransactionKind` is not really intended to be used directly by the end user, I agree that the proposed naming is more explicit.

src/eh1/delay.rs
dbrgn: It depends on the location where the text is displayed 🙂 As an example, the enum `ErrorKind` here is described using this format (first a short line with a summary, then a paragraph with more detailed explanations): https://docs.rs/embedded-hal/latest/embedded_hal/digital/enum.ErrorKind.html In the parent module, not the full text is displayed, but only the first paragraph: https://docs.rs/embedded-hal/latest/embedded_hal/digital/index.html As a consumer of docs, this helps a lot in my opinion to keep things tidy.

In the case of enum _variants_ I have to admit that right now there probably isn't an advantage in rustdoc. I did not realize that when writing my comment. However, I would still consider the pattern a best practice, just like with git commit messages: First line/paragraph is a short summary, optionally followed by further paragraphs that contain more details.

> Both of them look identical to me. I would suggest that you do your preferred way of formatting and just push directly to my fork.

If that's OK for you, I'll do that in this case!

dbrgn:
> I think there is a valid reason for doing it this way. (...)

You are right, thanks for your explanation!

dbrgn (approved):
I think this is now good to merge! Thanks @Artur-Romaniuk for your patience!

PR #103, opened 2023-12-26
Update to embedded-hal v1.0

Dirbaio:
- Update to embedded-hal v1.0
- wip patch, remove when v1.0 is actually released.

PR #105, opened 2024-01-09
update to `embedded-hal` v1.0.0 (no more RC!)

rursprung:
see the individual commit messages for details


Dirbaio:
dup of #103 :sweat_smile:

rursprung:
> dup of #103 😅

darn, i missed that! i did check for PRs but only checked the top one (because it was only released a few minutes ago) and didn't think that you'd play the long game and prepare this ahead of time!

since you also fix the doc links i think it'd be better if yours were merged (and the `nb` update can be done separately)

newAM:
Merged #103 :)

PR #106, opened 2024-01-09
CI: test against rustc 1.75 instead of nightly

rursprung:
nightly was only needed for some unstable features which have been stabilised in 1.75.

the alternative to this commit would be to just increase the MSRV of the crate to 1.75 (even though it's only 1.75 when using the async feature) and test against 1.75 only (instead of also testing against 1.63).


dbrgn:
Makes sense, thx!

dbrgn:
I'll publish the non-RC release today.

PR #107, opened 2024-01-10
update to `nb` 1.1

rursprung:
0.1.3 internally anyway already forwarded everything to 1, so there's no reason to depend on 0.1 anymore.


newAM (approved):
Thanks!

PR #108, opened 2024-01-10
add VS Code & JetBrains IDEA (RustRover, CLion, etc.) to `.gitignore`

rursprung:


dbrgn:
In general such ignores should be in your personal .gitignore file, if you use these IDEs, but it doesn't hurt to add them here either.

rursprung:
> In general such ignores should be in your personal .gitignore file, if you use these IDEs, but it doesn't hurt to add them here either.

that is a good point, thanks! i wonder why i've never set this up before 🤔
though i think a lot of people are unaware of it and thus having it in the repo for the most common IDEs (probably the two which i added?) doesn't hurt and is what i see in a lot of projects. thanks for merging!

and i've now added it to my global `~/.gitignore`. now i just need to remember to set up this config everywhere 😅

dbrgn:
Yeah, and there's also a downside: If you _do_ want to add a `.vscode/` directory (or similar) to a project, it will be ignored and must be un-ignored 🙂

Note that you can also [set up git configs conditionally, depending on path prefix](https://github.com/dbrgn/dotfiles/blob/d6bb3051ba878f0d1ddba8ec932313c31fcb415f/git/config#L43-L44).

Issue #112, opened 2024-01-31
Rename `pin` to `digital` to match `embedded-hal`


tommy-gilligan:
Currently we have `src/eh1/pin.rs` but I think this should be `src/eh1/digital.rs` to match `embedded-hal`.

dbrgn:
Fixed in #113.

PR #113, opened 2024-01-31
Rename pin to digital

tommy-gilligan:
https://github.com/dbrgn/embedded-hal-mock/issues/112

### DIFF
--- a/src/eh1.rs
+++ b/src/eh1.rs
@@ -8,8 +8,8 @@ mod error;
 pub use crate::eh1::error::MockError;
 
 pub mod delay;
+pub mod digital;
 pub mod i2c;
-pub mod pin;
 pub mod pwm;
 pub mod serial;
 pub mod spi;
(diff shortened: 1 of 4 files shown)


src/eh1.rs
```diff
@@ -8,8 +8,10 @@ mod error;
 pub use crate::eh1::error::MockError;
 
 pub mod delay;
+pub mod digital;
 pub mod i2c;
-pub mod pin;
+#[deprecated]
+pub use digital as pin;
```
dbrgn, line 14: Does this show up as a deprecation warning when importing `pin`?

If yes, could we add a message to the deprecation warning, indicating what to use instead? https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

dbrgn: I agree. We're still at 0.x, so a breaking change is fine.

dbrgn:
@tommy-gilligan I rebased your branch, squashed the commits and improved the commit message. Let's merge this as a breaking change. Thanks!

dbrgn (approved):

PR #119, opened 2024-05-22
Added async I2C implementation

### LINKED ISSUE #118: the trait `embedded_hal_async::i2c::I2c` is not implemented for `Generic<embedded_hal_mock::eh1::i2c::Transaction>`
Hello, thank you for this crate, it's awesome! I'm having an issue with mocking `embedded_hal_async::i2c::I2c`. Is there support for this?

asasine:
I saw that async I2C wasn't implemented and I'd like to use it for a sensor driver I'm writing. I implemented this following how `embedded_hal_mock::spi::Spi` was implemented for async.

Closes #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,17 @@ me know.
 
 ## Status
 
-- [x] I²C
-- [x] SPI
-- [x] No-op delay
-- [x] Actual delay
-- [x] Serial
-- [ ] RNG
-- [x] I/O pins (including PWM)
-- [x] ADC
-- [x] Timers (with `embedded-time` Cargo feature)
-- [ ] ...
+| Feature                                     | embedded-hal | embeded-hal-async |
+|---------------------------------------------|--------------|-------------------|
+| I²C                                         | ✅           | ✅               |
+| SPI                                         | ✅           | ✅               |
+| No-op delay                                 | ✅           | ✅               |
+| Actual delay                                | ✅           | ✅               |
+| Serial                                      | ✅           | -                |
+| RNG                                         | -            | -                |
+| I/O pins (including PWM)                    | ✅           | -                |
+| ADC                                         | ✅           | -                |
+| Timers (with `embedded-time` Cargo feature) | ✅           | -                |
 
 Pull requests for more mock implementations are welcome! :)
 
--- a/src/eh1/i2c.rs
+++ b/src/eh1/i2c.rs
@@ -298,6 +298,34 @@ impl i2c::I2c for Mock {
     }
 }
 
+#[cfg(feature = "embedded-hal-async")]
+impl embedded_hal_async::i2c::I2c for Mock {
+    async fn read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Self::Error> {
+        eh1::i2c::I2c::read(self, address, buffer)
+    }
+
+    async fn write(&mut self, address: u8, bytes: &[u8]) -> Result<(), Self::Error> {
+        eh1::i2c::I2c::write(self, address, bytes)
+    }
+
+    async fn write_read(
+        &mut self,
+        address: u8,
+        bytes: &[u8],
+        buffer: &mut [u8],
+    ) -> Result<(), Self::Error> {
+        eh1::i2c::I2c::write_read(self, address, bytes, buffer)
+    }
+
+    async fn transaction<'a>(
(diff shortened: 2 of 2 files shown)


asasine:
I don't love how the tests are implemented because there's a lot of repetition. I was trying to come up with a way to share the body between the sync and async tests, but I'm a bit too new to Rust to find a good way to do that. I was thinking something along these lines:

```rs
#[cfg(test)]
mod test {
    use super::*;

    async fn write<I /* : something? */>()
    {
        let expectations = [Transaction::write(0xaa, vec![10, 12])];
        let mut i2c = Mock::new(&expectations);
        I::write(&mut i2c, 0xaa, &vec![10, 12]).await.unwrap();
        I::done(&mut i2c,);
    }

    #[test]
    fn test_write() {
        use embedded_hal::i2c::I2c;
        write::<I2c>().await;
    }

    #[tokio::test]
    #[cfg(feature = "embedded-hal-async")]
    async fn test_write_async() {
        use embedded_hal_async::i2c::I2c;
        write::<I2c>().await;
    }

  // --snip--
}
```

felipebalbi:
I just tested this PR on a crate of my own and can confirm it works as expected. Thanks for doing the work @asasine.

asasine:
@dbrgn is there a finance your can review and release this? I believe there are a couple of unreleased fixes too that would be great to have available on crates.io.

README.md
```diff
@@ -39,16 +39,17 @@ me know.
...
+|---------------------------------------------|--------------|-------------------|
+| I²C                                         | ✅           | ✅               |
+| SPI                                         | ✅           | ✅               |
+| No-op delay                                 | ✅           | ✅               |
+| Actual delay                                | ✅           | ✅               |
+| Serial                                      | ✅           | -                |
+| RNG                                         | -            | -                |
+| I/O pins (including PWM)                    | ✅           | -                |
+| ADC                                         | ✅           | -                |
+| Timers (with `embedded-time` Cargo feature) | ✅           | -                |
```
dbrgn, line 52: 👍🏻

src/eh1/i2c.rs
```diff
@@ -298,6 +298,34 @@ impl i2c::I2c for Mock {
     }
 }
 
+#[cfg(feature = "embedded-hal-async")]
+impl embedded_hal_async::i2c::I2c for Mock {
+    async fn read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Self::Error> {
+        eh1::i2c::I2c::read(self, address, buffer)
```
dbrgn, line 304: The `eh1` import is aliased as `embedded_hal` further above, so this could be written like this:

```suggestion
        embedded_hal::i2c::I2c::read(self, address, buffer)
```

I like about this variant that it's clearly visible that the `embedded_hal_async` impl calls the `embedded_hal` trait impls.

Another option would be to directly import the `I2c` trait in the imports section, then you could write:

```suggestion
        I2c::read(self, address, buffer)
```

I like about that variant that it's clean and concise.

I tend towards the second option. What do you think? (Of course this would apply to all other usages of the trait as well.)

dbrgn (changes requested):
Nice, this looks quite straightforward!

Regarding the tests: If the async impls are doing nothing else than calling the sync impls, the additional tests don't add that much value and are mostly repeated code doing the same thing under the hood. Maybe one single big test that calls every async method (read, write, write_read, transaction) sequentially (plus very simple expectations for that) would suffice?

src/eh1/i2c.rs
asasine: I agree that the second option is clean and concise. The `use embedded_hal_async::i2c::I2c;` in the async tests would remain necessary as the call to i.e. `write(...)` must explicitly invoke the asynchronous version.

asasine:
I left the tests as-is for now since the same question exists for all other async implementations that are async-over-sync (delay and SPI) and it would best (in my opinion) to address all of them in one pass.

dbrgn (approved):
Thanks! 👍🏻

dbrgn:
> I left the tests as-is for now since the same question exists for all other async implementations that are async-over-sync (delay and SPI) and it would best (in my opinion) to address all of them in one pass.

Fair point.

asasine:
No problem, thanks for the quick turnaround!

PR #120, opened 2024-05-26
Remove redundant async tests, do some cleanup

dbrgn:
Some cleanup commits. The async simplification was discussed here: https://github.com/dbrgn/embedded-hal-mock/pull/119#pullrequestreview-2079775113 @asasine do you want to take a look at this PR?

Note: The delay async tests were not removed, since in the delay module, the async impl doesn't just call through to the sync impl.


asasine (approved):
Looks good to me!

From the perspective of the test, there's no way to know that the impl is async-over-sync, so it would be great to come up with a way to actually run through the same test body for both, but I agree with your earlier point that of probably doesn't warrant the complexity.

PR #121, opened 2024-08-02
Fix checking address on I2C transaction start/end

eldruin:
I have noticed that the addresses supplied in the transaction start/end are not checked. Hence this PR.
Although I am wondering now if the transaction start/end should have an address parameter at all. Maybe we should remove it altogether.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Check address in I2C transaction start/end operations. (#121)
+
 ### Changed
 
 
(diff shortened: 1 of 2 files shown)


dbrgn:
> Although I am wondering now if the transaction start/end should have an address parameter at all. Maybe we should remove it altogether.

How would you pass the address argument to the read/write functions in that case?

eldruin:
The address is already checked in the individual steps.
```rust
[ I2cTrans::transaction_start(ADDR),
  I2cTrans::write(ADDR, vec![0]),
  I2cTrans::read(ADDR, &mut data),
  I2cTrans::transaction_end(ADDR) ]
```
vs.
```rust
[ I2cTrans::transaction_start(),
  I2cTrans::write(ADDR, vec![0]),
  I2cTrans::read(ADDR, &mut data),
  I2cTrans::transaction_end() ]
```

asasine:
Removing the address from the transaction start is opposite to how the underlying bus works for multiple operations. The address is sent as part of start/repeated start, and not as part of the write/read operations. If anything, the address should be only included in the starts and removed from everything else.

PR #122, opened 2024-09-09
remove `eh0` feature from default features

rursprung:
`embedded-hal` v1.0 came out in january 2024, going forward new code should use this. accordingly, `eh0` support should no longer be part of the default features to reduce the amount of dependencies being pulled in.

existing code will likely still use it, thus consumers will have to explicitly enable the feature when upgrading.

CC @tommy-gilligan (you suggested the same [as part of your larger draft](https://github.com/dbrgn/embedded-hal-mock/pull/115/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R29)) & @eldruin (i noticed it due to your comment here: https://github.com/rust-embedded-community/tb6612fng-rs/pull/51#pullrequestreview-2154759310)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Drop fixed MSRV policy (#124)
+- **Breaking**: the `eh0` feature is no longer part of the default features.
+  it still exists as an optional feature and has to be explicitly added when needed.
 
 
 ## 0.11.1 - 2024-06-02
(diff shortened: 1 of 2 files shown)


dbrgn:
I agree with this change 🙂 Let's wait until #124 is merged.

dbrgn:
@rursprung merged, you can rebase now!

2bndy5:
FYI, the README still states that the `eh0` feature is enabled by default.

https://github.com/dbrgn/embedded-hal-mock/blob/e11348e9d34ef5809eb3b13979bab0efd7d92a44/README.md?plain=1#L29

newAM:
> FYI, the README still states that the `eh0` feature is enabled by default.

Thanks!  Fixed in #140

2bndy5:
> > FYI, the README still states that the `eh0` feature is enabled by default.
> 
> Thanks! Fixed in #140

Oops, you mean #130. Ticket 140 doesn't exist (yet).

PR #124, opened 2024-09-13
Drop strict MSRV policy

### LINKED ISSUE #123: MSRV check broken
the CI tests on rust 1.63 to ensure MSRV but it doesn't use a `Cargo.lock`. since not all crates have strict MSRV policies it can happen that a newer version of a dependency is being pulled in which breaks the CI for MSRV.

this is exactly what just happened for #122: `error: package `tokio-macros v2.4.0` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.63.0` (see [the full build output](https://github.com/dbrgn/embedded-hal-mock/actions/runs/10770307849/job/29863392015?pr=122))

there are three ways to deal with this:
* slacken the MSRV policy ([`defmt` style](https://github.com/knurling-rs/defmt?tab=readme-ov-file#msrv))
* check in `Cargo.lock` (see [the cargo FAQ](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control) & [blog post](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html))
* check in a copy of `Cargo.lock` specific for the MSRV check (e.g. as `Cargo.lock.msrv`) and copy it to `Cargo.lock` as part of the CI script (downside: it has to be manually maintained if dependencies are added/updated by copying over the correct `Cargo.lock` to `Cargo.lock.msrv` to commit it)

dbrgn:
See #123. Fixes #123.

Note: We still use a MSRV in CI (to avoid accidentally increasing the MSRV without a need), but we can bump it at any time without a breaking semver release (e.g. when a transitive dependency drops support for our current MSRV).


rursprung (approved):

eldruin (approved):
Thank you for your work!

newAM (approved):

PR #125, opened 2025-01-24
Added WaitForEdgeForever and WaitForStateForever transactions, which return a Pending future to the callee.

leighleighleigh:
I have some code I'd like to mock for a quadrature encoder, which uses `select(pin_clk,pin_data)`. 
In order to properly test it, I need to be able to mock different paths in the select statement.

As it stands currently, the current async transactions always return `Ok()` when called - this means my `select()` paths are always taking the first branch.

The solution I propose here is to add a modified `WaitForEdge/WaitForState` enum called `WaitFor{State,Edge}Forever`, which returns a `Pending` future. This simulates a pin that never changes, while still allowing a mock transaction for the actual `wait_for_edge` call.

I have added tests for this, which poll the returned future for 10 milliseconds before timing-out (successful result).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ eh0 = ["dep:eh0", "dep:nb"]
 eh1 = ["dep:eh1", "dep:embedded-hal-nb"]
 
 embedded-time = ["dep:embedded-time", "dep:void"]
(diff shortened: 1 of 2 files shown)


ryankurte (approved):
seems good to me!

dbrgn:
Thanks for the review, @ryankurte!

PR #130, opened 2025-04-25
README.md: fix default features

newAM:
Ref: https://github.com/dbrgn/embedded-hal-mock/pull/122#issuecomment-2831037346


dbrgn:
Good catch.

PR #132, opened 2025-04-25
refactor!: `eh1::digital` use `embedded_hal::digital::ErrorKind`

### LINKED ISSUE #129: use  `embedded_hal::digital::ErrorKind` for mocked digital pins
[embedded-hal-errors]: https://docs.rs/embedded-hal/1.0.0/embedded_hal/digital/enum.ErrorKind.html?search=ErrorKind

I find it weird that `eh1::MockError` only has an `Io` variant for an `std::io::ErrorKind`. Because most embedded rust cannot use `std::io` in production code, the tested scenario is unlikely to ever occur. IMHO, it would make more sense to allow `eh1::digital::Transaction::with_error()` to accept `embedded_hal::digital::ErrorKind` parameter type. This should be especially useful for measuring coverage of an embedded driver lib that wraps [`embedded_hal::*::ErrorKind`][embedded-hal-errors] into a custom Error type.

Would it be feasible to let `eh1::digital::Transaction::with_error()` accept `embedded_hal::digital::ErrorKind` parameter type?

> [!caution]
> This would break backward compatibility downstream because `MockError` is a different type. Adding more variants to the `MockError` enum wouldn't cover code that (for example)
>
> ```rs
> use embedded_hal::digital::ErrorKind as OutputPinError;
> 
> impl From<OutputPinError> for CustomError<OutputPinError> {
>     fn from(value: OutputPinError) -> Self {
>         CustomError::DigitalOutError(value)
>     }
> }
> ```

2bndy5:
resolves #129 

Please review diligently as I'm not overly familiar with this crate's code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Drop fixed MSRV policy (#124)
 - **Breaking**: the `eh0` feature is no longer part of the default features.
   it still exists as an optional feature and has to be explicitly added when needed.
+- **Breaking**: Switch to `embedded_hal::digital::ErrorKind` (replaces previous type `eh1::MockError`)
+  for `eh1::digital::Transactions`. (#132)
 
 
 ## 0.11.1 - 2024-06-02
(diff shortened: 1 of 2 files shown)


ryankurte (approved):
lgtm, thanks!

PR #142, opened 2026-06-29
[codex] add optional mock labels

### LINKED ISSUE #43: Feature Request: Add Optional labels to Mocks
First of all, thanks for your work here. This project has been a real time-saver.

When you have a large set of mocks, it can become quite difficult to debug a test failure.

Say for example that you have a test that mocks 4 GPIO pins, with a specific sequence of reads and writes. As it currently stands you might get a message like this;
```
---- context::tests::send_command stdout ----
thread 'context::tests::send_command' panicked at 'no expectation for pin::is_high call', /usr/local/cargo/git/checkouts/embedded-hal-mock-54d083840a293f52/d394c81/src/pin.rs:159:50
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Apart from the offending function call, this error message doesn't yield much useful information to track down where the failure has occurred. To improve these error messages I'd like to add a label so that;
` 'no expectation for pin::is_high call'` becomes ` 'no expectation for pin::is_high call on pin my_label'`

From the user facing API this might look something like;
```rust
// Not labelled
let mut pin = PinMock::new(&expectations);

// Labelled
let mut pin = PinMock::new_with_label(&expectations, "my_label");
```

I'm happy to have a crack at implementing this, but I thought I get some input before going ahead with it.

Old-Ding:
## Summary

Closes #43.

This adds optional labels to the generic mock state so missing-expectation failures can identify which mock instance failed when a test uses several mocks at once. Unlabeled mocks retain their existing panic messages.

## Changes

- Added `Generic::with_label(expected, label)` while keeping `Generic::new(expected)` unchanged.
- Added a shared `next_expectation(message)` helper that appends the label to the existing missing-expectation message when present.
- Routed every Generic-backed expectation path in ADC, I2C, digital, delay, PWM, and SPI mocks through the shared helper.
- Added regression tests for labeled and unlabeled messages, public type-alias usage, ADC, and both embedded-hal I2C generations.

## Validation

Local checks passed:

- `cargo test --all-features` (145 unit tests and 21 doc tests)
- `cargo +1.75.0 test --all-features` (145 unit tests and 21 doc tests)
- `cargo fmt --all -- --check`
- `git diff --check`


newAM (changes requested):
`new_with_label` should be `with_label` to match naming idioms (such as `Vec::new` / `Vec::with_capacity`) but it otherwise looks good to me.

Old-Ding:
Addressed on the current head (`303db16`): the public constructor is now `Generic::with_label`, and no `new_with_label` references remain. This follows the naming convention you noted.

Verification on the current head:
- `cargo test --all-features`: 145 unit tests and 21 doc tests passed.
- `git diff --check upstream/main...HEAD`: clean.

PR #143, opened 2026-07-05
Return errors from I2C transaction operations

### LINKED ISSUE #127: Unwrap in I2C transaction
Let's say I have an I2C transaction like this one below. Clearly, any of the operations can fail and in this case error should be returned (At least `I2c` interface allows this). To simulate such failed operation, I was expecting `with_error` approach to work (Line 3).
```rs
let mut i2c = Mock::new(&[
    I2cTransaction::transaction_start(0x76),
    I2cTransaction::write(0x76, vec![0x88]), // .with_error(I2cErrorKind::Other),
    I2cTransaction::read(0x76, vec![0xCC, 0x6D, 0x1B, 0x68, 0x32, 0x00]),
    I2cTransaction::write(0x76, vec![0xA1]),
    I2cTransaction::read(0x76, vec![0x4B]),
    I2cTransaction::write(0x76, vec![0xE1]),
    I2cTransaction::read(0x76, vec![0x70, 0x01, 0x00, 0x13, 0x21, 0x03, 0x1E]),
    I2cTransaction::write(0x76, vec![0xFA]),
    I2cTransaction::read(0x76, vec![0x7E, 0x9B, 0x00, 0x5D, 0xC2])
        .with_error(I2cErrorKind::Other),
    I2cTransaction::transaction_end(0x76),
])
```
But it does not and causes panic instead. This panic originates from `unwrap` on the following line:
https://github.com/dbrgn/embedded-hal-mock/blob/e11348e9d34ef5809eb3b13979bab0efd7d92a44/src/eh1/i2c.rs#L273
For me it seems `?` would be more appropriate.

Old-Ding:
Fixes #127.

Summary:
- Stop unwrapping nested I2C operation results inside `eh1::i2c::Mock::transaction`.
- Return the first mocked operation error to the caller.
- Still consume the `transaction_end` expectation so the mock transaction boundary remains balanced when an operation returns an error.
- Add regression coverage for both the returned operation error and the unhappy path where extra expectations appear before `transaction_end`.

Testing:
- `cargo test transaction_returns_operation_error --features eh1`
- `cargo test transaction_error_rejects_trailing_expectations --features eh1`
- `cargo fmt --check`
- `cargo test --all-features`
- `git diff --check HEAD~1..HEAD`
- `git ls-files --eol src/eh1/i2c.rs`


dbrgn:
@Old-Ding If you're copy-pasting directly from AI output, please ensure at the very least that the description is readable and well-formatted...

src/eh1/i2c.rs
```diff
@@ -447,6 +452,23 @@ mod test {
         i2c.done();
     }
 
+    #[test]
+    fn transaction_returns_operation_error() {
```
dbrgn, line 456: This tests the happy path. Could you add a test for the unhappy path as well, where the user included additional assertions before `transaction_end`? This should cause the test to fail.

Probably something like this:

```
#[test]
#[should_panic(expected = "i2c::transaction_end unexpected mode")]
fn transaction_error_rejects_trailing_expectations() {
    let expectations = [
        Transaction::transaction_start(0x76),
        Transaction::write(0x76, vec![0x88]).with_error(ErrorKind::Other),
        Transaction::write(0x76, vec![0xAA]),
        Transaction::transaction_end(0x76),
    ];
    let mut i2c = Mock::new(&expectations);
    let bytes = [0x88];
    let more = [0xAA];
    let mut ops = [i2c::Operation::Write(&bytes), i2c::Operation::Write(&more)];
    let _ = i2c.transaction(0x76, &mut ops);
}
```

...but I did not test this in practice, so please validate that it works as intended.

dbrgn (changes requested):
Regarding the change itself: I was a bit on the fence, whether a `TransactionEnd` expectation should be part of the expectations (since the transaction is aborted, not completed). But on the wire a STOP happens as well, so `TransactionEnd` is probably correct.

Having more expectations between error and `TransactionEnd` would be wrong, and that will get caught by this crate. This is not yet tested, so an additional test would be nice (see inline comment).

src/eh1/i2c.rs
Old-Ding: Addressed in `dbc0ede`: `transaction_error_rejects_trailing_expectations` adds the error-path case with an extra `write` expectation before `transaction_end`, and asserts that it panics with `i2c::transaction_end unexpected mode`.

I validated the current head with `cargo test --all-features`: 140 unit tests and 21 doc tests passed, including this regression test.

Issue #135, opened 2025-10-22
Support for don't care writes in I2C


simonsso:
### Motivation
I have a test case where there are a ton of I²C writes which I don't care about in this specific test. I still want to mock execute them but the data written is not relevant.

### Implementation
I have some ideas how this could be implemented. 

The `impl Transaction` block [1] could be extended with a function `write_ignore` then either implemented by extending the enum  `Mode`[2]   with `WriteIgnore`  or a new field in the `Transaction` struct[3]

* Are there any negative implications I did not consider?
* Any preferred implementation style?

[1]: https://github.com/dbrgn/embedded-hal-mock/blob/18b678a8eec1e33141167346e790af0a94fd9e94/src/eh1/i2c.rs#L95
[2]: https://github.com/dbrgn/embedded-hal-mock/blob/18b678a8eec1e33141167346e790af0a94fd9e94/src/eh1/i2c.rs#L66
[3]: https://github.com/dbrgn/embedded-hal-mock/blob/18b678a8eec1e33141167346e790af0a94fd9e94/src/eh1/i2c.rs#L83

simonsso:
@ryankurte thanks for merging #136 As soon as there is a release on creates.io we could close this ticket too!

PR #144, opened 2026-07-05
Validate I2C addresses with ignored writes

Old-Ding:
## Summary
- keep I2C address validation active when `with_ignore_write()` skips write-data comparison
- keep `write_read` response length validation active so mismatches report the mock assertion before copying
- add regression coverage for ignored-write address and response-length mismatches

## Testing
- `cargo fmt --check`
- `cargo test --all-features`
- `git diff --check`


dbrgn:
@simonsso @2bndy5 you were both involved in #135 - what would your expectation be when using `with_ignore_write`? Would you expect the address to be checked or not?

(Usually this would be a breaking change, but in this case it was not released on crates.io so far.)

2bndy5:
Concerning #135, I will defer to @simonsso. But it would make sense to me if the address is checked; we wouldn't want production code sending data to the wrong I2C device.

simonsso:
@dbrgn sorry for the delay I have been on a long vacation and is just catching up.

When I needed this I was just ignoring everything, but thinking about it it again now... It makes sense to check the address even if the contents of the rest of the write is ignored.

simonsso (approved):
Looks reasonably

